PR #62, opened 2014-06-03
When parsing the generate json, scenario outline elements are ignored

tcoppotelli:
This will fix the following issue:
https://github.com/masterthought/jenkins-cucumber-jvm-reports-plugin-java/issues/96


damianszczepanik:
This problem is fixed in damianszczepanik/cucumber-reporting/pull/179

Also I found that attached test file is wrong because there is no 'id' tag in https://github.com/tcoppotelli/cucumber-reporting/blob/scenarioOutlineFix/src/test/resources/net/masterthought/cucumber/scenario_outline.json#L19

PR #85, opened 2015-03-29
Updated example in README.md to make it ready for copy&paste

damianszczepanik:


damianszczepanik:
I've started developing with cucumber-reporting so once I launched HelloReport I've updated example a little bit.

Issue #76, opened 2014-11-04
Display Undefined Steps


chanceVermilion:
Cucumber will mark steps as "undefined" if there is no definition for them, but the reporting does not include this number which leads to some confusing reports like in the attached image.
![screen shot 2014-11-04 at 17 07 00](https://cloud.githubusercontent.com/assets/2787797/4908971/3c9253ec-646f-11e4-8464-f610e1477d36.png)

Adding in a column for "undefined" steps would fix this issue

damianszczepanik:
Support for undefined and missing steps have been added in #92.

Issue can be closed.

Issue #86, opened 2015-04-06
101% tests were executed


damianszczepanik:
Having 2076 tests passed and 51 failed I can observe results with 97% pass and 4% failed what means I have executed 101% tests :)

Will try to find out whether this is plugin problem or cucumber-reporting.

kingsleyh:
I guess it's the reporting - could be miscounting scenario outline or backgrounds 

If you don't mind maybe you could send me your cucumber json to kingsley at Masterthought dot net so I can investigate

damianszczepanik:
I could not reproduce it without json files which I did not store... If it happens again I will handle this issue.

msymons:
I am seeing this issue currently with Jenkins plugin v0.0.23.  I can provide the json files if it would help.  Just let me know where to send them (as I cannot attach them to a public issue tracker).

damianszczepanik:
It would be. Tou can share by attaching the file or sharing by gist or github. Io

damianszczepanik:
Should not happen after chart library upgrade

PR #92, opened 2015-05-02
- Reduced number of generic Exceptions

damianszczepanik:
- Merged duplicated code into loops
- Reduced code complexity, increased scalability
- Moved Status to public type (separate file)
- Linked page footer to github project
- Replaces number of counters with Map that keeps all values in single instance and allow to extend for new one
- Updated tests so they don't mix skipped and pending results
- Updated colors so Skipped/Pending/Undefined/Missing use the same yellow background color
- each chart presents all kind of results (not only pass/fail)
- fixed colors so they are the same on each page and chart
- updated some tests to make them deterministic and simpler

Issue #76

### DIFF
--- a/src/main/java/net/masterthought/cucumber/Background.java
+++ b/src/main/java/net/masterthought/cucumber/Background.java
@@ -1,31 +1,27 @@
 package net.masterthought.cucumber;
 
+import java.util.HashMap;
+import java.util.Map;
+
+import net.masterthought.cucumber.util.Status;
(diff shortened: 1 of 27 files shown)


kingsleyh:
if this could be automatically merged it would help as I have no time to investigate this

damianszczepanik:
@kingsleyh 

PR updated,
can you also review issues and close those which are already fixed? I will take a look and fix some of them including #86.

Issue #49, opened 2014-01-10
Gherkin syntax is not properly HTML escaped


btaz:
When the Gherkin scenario contains e.g. XML, then it's not properly escaped when displayed on the cucumber-reporting HTML page.

![screen shot 2014-01-09 at 10 00 08 pm](https://f.cloud.github.com/assets/597495/1885066/3803ac7e-79bd-11e3-96b3-665933a882b6.png)

This unfortunately displays as:

![screen shot 2014-01-09 at 9 58 48 pm](https://f.cloud.github.com/assets/597495/1885068/55519b56-79bd-11e3-8b3c-94e137f67ccb.png)

The problem is that the browser tries to parse the XML tags as HTML. Any of these characters need to be escaped: & < > " /

One option is to use velocity-tools EscapeText for this.
http://velocity.apache.org/tools/devel/generic/EscapeTool.html

I created a pull request for this:
https://github.com/masterthought/cucumber-reporting/pull/48

SlyAtBest:
I've experienced a similar issue when using an Assert.assertEquals() to test a boolean value.  The value is returned as either &lt;true> or &lt;false>, which is not shown in the generated report when rendered by a browser.  I'm guessing this is because it is being treated as a html tag and so this needs to be escaped in order for it to display properly.  The page source reveals that the underlying values do exist.

Example:
> boolean expectedType = true;
> boolean currentType = false; 

> Assert.assertEquals("Package is not listed as the correct type",
                expectedType, currentType);

Should produce:

> Package is not listed as the correct type expected:&lt;true> but was:&lt;false>

But instead produces:

> Package is not listed as the correct type expected:<true> but was:<false>

damianszczepanik:
This issue has been already resolved and released.

msymons:
In what version was this resolved and released?  The release notes only go up to v0.0.21:

https://github.com/masterthought/cucumber-reporting/wiki/Release-Notes

The same is also true for jenkins-cucumber-jvm-reports-plugin-java  (where I have developers refusing to upgrade the plugin until I can tell them what has changed).

damianszczepanik:
New cucumber reporting has been released weeks ago. New version is available also via Jenkina

msymons:
[...]

Issue #84, opened 2015-02-23
Release components


fmodesto:
Could you please release the component with the latest fixes, and also update and release the maven and jenkins plugins.

Thank you,

damianszczepanik:
New release has been published

PR #96, opened 2015-05-03
Gherkin syntax is not properly HTML escaped

damianszczepanik:
Issue #49

### DIFF
--- a/src/main/java/net/masterthought/cucumber/json/Element.java
+++ b/src/main/java/net/masterthought/cucumber/json/Element.java
@@ -9,6 +9,7 @@ import net.masterthought.cucumber.ConfigurationOptions;
 import net.masterthought.cucumber.util.Status;
 import net.masterthought.cucumber.util.Util;
 
+import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.commons.lang.StringUtils;
 
 import com.googlecode.totallylazy.Function1;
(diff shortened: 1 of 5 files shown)


kingsleyh:
thanks for your contributions and help damian - you are AWESOME! As you can see I have had zero time to work on this project since having my baby daughter 2 years ago!

damianszczepanik:
Hope you have some time soon to make new release since this is why I'm preparing PRs for this and that's what other users expect (#84).

We are using cucumber reports everyday so I have some idea how to improve it and make more and more friendly!

kingsleyh:
good job :)

I can make a new release on the following condition:

 -- The master branch of cucumber reports works with the jenkins cucumber plugin, the maven mojo and cucumber-sandwich on master and slave without any new issues or major problems.

This is because I have no time to regression test this which is one of the major reasons I don't manage to release often.

Let me know approx when you want a new release and I will try to be prepared for it. 

Thanks

--K

damianszczepanik:
Sounds good. I should be ready within 3-4 weeks.

If you advice me how to build binary which I can test locally on Jenkins that would be great.

kingsleyh:
build cucumber reports locally - and then in the jenkins project pom change the version of cucumber reports to the one you built locally and then run mvn clean install and it should produce an hpi in the target directory - use that hpi in a locally running jenkins

damianszczepanik:
mvn install produced jar file not hpi - are you sure about the command?

I can't figure out what is the goal based on information from pom.xml file so your help is welcome :-)

PR #99, opened 2015-05-08
Updated duration format

damianszczepanik:
- leading 0 are added to ms, s and h to be able to easily compare visually two durations
- duration is align to the right
- used <th> instead of <td> + bold


damianszczepanik:
@kingsleyh 
I have built and upgraded plugin after recent changes. I've created report with old version, then upgrade plugin (hpi) and launch another build to check if there is any problem and looks nice!

The question is: now you can configure pending and unresolved statuses as failed but now that would be nice to add the same option for missing and undefined. Any suggestion how make it work to keep backwards compatibility? I would have to add two more input parameters into ReportBuilder constructor...

kingsleyh:
Hi great,

You could add another constructor and in the existing constructor set the new items to a default inside.

If we change the existing constructor we would have to make sure it works in Jenkins plugin, cucumber sandwich and mojo and if anyone is using the library directly.

Another option as well so Could also add a map as the new param to make adding new Params easier

--k

PR #100, opened 2015-05-10
pending and missing status can be defined as build failure during Rep…

damianszczepanik:
…ortBuilder creation


damianszczepanik:
@kingsleyh 
conflict resolved, locally tested with all PRs you've already merged - can you estimate how long does it take to have new release available over jenkins repo?

kingsleyh:
When I release I push the following to nexus:
1. cucumber-reporting
2. maven mojo
3. cucumber-sandwich
then I release to jenkins repo - which takes a few mins - but only gets updated on the remote every 6 hours. So the max time it would take in theory is 6 hours.

The cucumber-reporting changes must not break maven mojo or cucumber-sandwich and of course the jenkins plugin.

Do you want me to make a new release with the current code in master?

--K

damianszczepanik:
I found one problem with jenkins plugin but fortunately this is already fixed but not released!

Consider merging https://github.com/masterthought/cucumber-reporting/pull/102 - since I've added support for pending and missing I think that is good idea to make major release.

Once you release new version i will update my jenkins from repo and wait for the feedback from users if there is anything they expect to have (to fix, to add).

Let's go. The last pull you should merge is https://github.com/masterthought/jenkins-cucumber-jvm-reports-plugin-java/pull/128 which updates the plugin to use new features.

PR #101, opened 2015-05-13
fixed cucumberjs report displaying After hook as being a step

BogdanLivadariu:
cucumberjs generates an extra attribute in the step json, causing the HTML report to display that step, and to look 'ugly'
```json
          {
            "keyword": "After ",
            "hidden": true,
            "result": {
              "duration": 1456494,
              "status": "passed"
            },
            "match": {}
          }
```
after step shown

with the current impl, the step is being hidden, and if that step has attachments, they will remain visible.
after2

after3
after4

### DIFF
--- a/src/main/java/net/masterthought/cucumber/util/Status.java
+++ b/src/main/java/net/masterthought/cucumber/util/Status.java
@@ -13,7 +13,8 @@ public enum Status {
     SKIPPED("#88AAFF"),
     PENDING("#FBB907"),
     UNDEFINED("#FBB957"),
-    MISSING("#FBB9A7");
+    MISSING("#FBB9A7"),
+    HIDDEN("");
 
     /** Color representation for given status. */
     public final String color;
--- a/src/main/resources/styles/reporting.css
+++ b/src/main/resources/styles/reporting.css
@@ -57,6 +57,10 @@
 	background-color: #C5D88A;
 }
 
+.hidden {
+	visibility: hidden;
+}
+
 .failed {
 	background-color: #D88A8A;
 }
--- a/src/main/java/net/masterthought/cucumber/json/Step.java
+++ b/src/main/java/net/masterthought/cucumber/json/Step.java
@@ -21,6 +21,7 @@ public class Step {
 
     private String name;
     private String keyword;
+    private boolean hidden;
     private String line;
     private Result result;
     private Row[] rows;
@@ -31,6 +32,10 @@ public class Step {
 
     public Step() {
 
+    }
+    
+    public boolean isHidden() {
+        return hidden;
     }
 
     public DocString getDocString() {
@@ -72,10 +77,12 @@ public class Step {
     }
 
     public Status getStatus() {
-        if (result == null) {
+        if (result == null && !isHidden()) {
             System.out.println("[WARNING] Line " + line + " : " + "Step is missing Result: " + keyword + " : " + name);
             return Status.MISSING;
-        } else {
+        } else if (isHidden()) {
+            return Status.HIDDEN;
+        }else {
             return Status.valueOf(result.getStatus().toUpperCase());
         }
     }


src/main/java/net/masterthought/cucumber/util/Status.java
```diff
@@ -13,7 +13,8 @@
     SKIPPED("#88AAFF"),
     PENDING("#FBB907"),
     UNDEFINED("#FBB957"),
-    MISSING("#FBB9A7");
+    MISSING("#FBB9A7"),
+    HIDDEN("");
```
damianszczepanik: HIDDEN is not valid value of status property so it should not be added here

src/main/resources/styles/reporting.css
```diff
@@ -57,6 +57,10 @@
 	background-color: #C5D88A;
 }
 
+.hidden {
+	visibility: hidden;
```
damianszczepanik: where this class is used?

BogdanLivadariu: When a step will have the hidden attribute in Json set to true then the
container div will have the class set to true, thus overriding the status
passsed or failed with hidden, making that step to be hidden from the
report since the display of that step would be redundant, as you can note
from the screenshots I have attached.

src/main/java/net/masterthought/cucumber/util/Status.java
BogdanLivadariu: I can't think of any other place this might go, since it is being used to
override any status of a step if the attribute "hidden" is to be found true

damianszczepanik: check with isBackground method so I guess you can make it similar but for sure not here

src/main/resources/styles/reporting.css
damianszczepanik: That sounds clear but I can't see in this PR where this class is enabled or disabled. Is it new class or re-declared from existing one?

src/main/java/net/masterthought/cucumber/util/Status.java
kingsleyh: At the moment the Java code contains logic and view concerns mixed together - long term I would like to separate these out so the Java code is view / display agnostic

BogdanLivadariu: This will require a major refactor :)

The view should be handled by the velocity template and the logic by the
java, so we would strip the java methods that wraps html code.

I think I can come up with something for 0.0.26 release.

src/main/resources/styles/reporting.css
BogdanLivadariu: There is no need for toggle, the passed,failed,pending, missing class will
be replaced with hidden only if the isHidden function is found true, as
there is no need to show something that is marked to be hidden.

In src/main/resources/styles/reporting.css
<https://github.com/masterthought/cucumber-reporting/pull/101#discussion_r30266049>
:

> @@ -57,6 +57,10 @@
>  	background-color: #C5D88A;
>  }
>
> +.hidden {
> +	visibility: hidden;

That sounds clear but I can't see in this PR where this class is enabled or
disabled. Is it new class or re-declared from existing one?

—
Reply to this email directly or view it on GitHub
<https://github.com/masterthought/cucumber-reporting/pull/101/files#r30266049>
.

src/main/java/net/masterthought/cucumber/util/Status.java
kingsleyh: I want to get rid of velocity as well - something modem - maybe handlebars

BogdanLivadariu: Sounds like a challenge:) can you put some thoughts in an issue, so we
would have some guidelines ?

BogdanLivadariu:
closing PR, since this just adds more VIEW noise inside java LOGIC

paymand:
@damianszczepanik: I'm facing the exact same issue with the hidden steps displayed as 'null' lines.
Could you fix it so that the hidden steps are not displayed?

image

damianszczepanik:
Try with the recent snapshot where hooks have been improved. If this does not work, please file new issue and attach json file.

paymand:
If you are referring to https://github.com/damianszczepanik/cucumber-reporting/commit/7030d0de4a5badeb311e382e42600cfcd9b3a158, I'm already on https://github.com/damianszczepanik/cucumber-reporting/commit/cfa2e993e10b06812dab69b1883b758d70d53478 which is after that.
Filed a new issue here: https://github.com/damianszczepanik/cucumber-reporting/issues/263

PR #125, opened 2015-07-28
Wrapping step error message in div.

tommywo:
I've added css class for wrapping error message in a parent div.

Before:
before

After:
after


damianszczepanik:
Well done :)

damianszczepanik:
Nice, thanks!

PR #127, opened 2015-08-03
Routing  build to container-based infrastructure

tommywo:
As described here:  http://docs.travis-ci.com/user/workers/container-based-infrastructure/,
build will startup faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,1 +1,2 @@
 language: java
+sudo: false


damianszczepanik:
I don't like experiments with sudo but if you follow travis guide...

Thanks for PR :)

damianszczepanik:
I had to revert this change since it breaks coverage tests - will investigate it later: should work your solution in theory but doesn't in practice.

PR #486, opened 2016-06-26
Adds ability to exclude tags from the 'Tags Overview' chart

aleitamar:
The motivation behind this is to allow tagging Cucumber scenarios with
JIRA issue IDs, while keeping the tags overview chart clean from
clutter.


damianszczepanik:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/damianszczepanik/cucumber-reporting/pull/486'>Refresh</a>.<a href='#crh-update-none'></a>

src/main/java/net/masterthought/cucumber/Configuration.java
```diff
@@ -94,4 +97,19 @@ public File getEmbeddingDirectory() {
...
+     */
+	public Pattern[] getTagsToExcludeFromChart() {
+		return tagsToExcludeFromChart;
+	}
+
+	/**
+	 * Stores the regex patterns to be used for filtering out tags from the 'Tags Overview' chart
+	 * @param patterns Regex patterns to match against tags
+	 */
+	public void setTagsToExcludeFromChart(Pattern... patterns) {
```
damianszczepanik: so it means that at the same time you can see filtered tags xor all tags? so once you want to change this, report needs to be rebuilt?

src/main/java/net/masterthought/cucumber/generators/TagsOverviewPage.java
```diff
@@ -50,6 +51,26 @@ static String generateTagLabels(List<TagObject> tagsObjectList) {
         return "[" + StringUtils.join(tagNames, ",") + "]";
     }
 
+	private List<TagObject> filterExcludedTags(List<TagObject> tagsObjectList) {
+		ArrayList<TagObject> filteredTags = new ArrayList<>();
```
damianszczepanik: change type to interface, not impelmentation

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
```diff
@@ -140,4 +141,33 @@ public void getProjectName_ReturnsProjectName() {
         // then
         assertThat(name).isEqualTo(projectName);
     }
+    
+    @Test
+	public void getTagsToExcludeFromChart_defaultConfiguration_ReturnsEmptyList() throws Exception {
```
damianszczepanik: defaultConfiguration -> could be removed, I use it only for no-default cases

damianszczepanik: does it have to declare to throw an exception?

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
```diff
@@ -140,4 +141,33 @@ public void getProjectName_ReturnsProjectName() {
...
+
+        // then
+        assertThat(patterns).isEmpty();
+	}
+    
+    @Test
+	public void getTagsToExcludeFromChart_addPatterns_ReturnsListWithAllPatterns() throws Exception {
+        // give
+        Configuration configuration = new Configuration(outputDirectory, projectName);
+        Pattern somePattern = Pattern.compile("");
```
damianszczepanik: StringUtils.EMPTY :)

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
```diff
@@ -140,4 +141,33 @@ public void getProjectName_ReturnsProjectName() {
...
+        Pattern anotherPattern = Pattern.compile("");
```
damianszczepanik: so what are you excluding now? This case is not fully clear to me, maybe rename variables and provide more meangful example?

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
```diff
@@ -140,4 +141,33 @@ public void getProjectName_ReturnsProjectName() {
...
+        configuration.setTagsToExcludeFromChart(somePattern, anotherPattern);
+        
+        // when
+        Pattern[] patterns = configuration.getTagsToExcludeFromChart();
+
+        // then
+        assertThat(patterns)
+        	.hasSize(2)
```
damianszczepanik: instead you can use containsExactly() or containsOnly

src/main/java/net/masterthought/cucumber/Configuration.java
damianszczepanik: also can you accept String[] so it would be easier to pass regexpt from cucumber Jenkins plugin

aleitamar: 1st comment:
yes, the report will need to be rebuilt if you wanted to change the filtered tags.
the filter only affects the chart, so you still have the full information in the table below the chart.

2nd comment:
i can accept String[] but then i need to handle invalid regex.
what do you think is the best way to deal with an invalid input in this case?
maybe i can add an overload to the configuration setter that will have String[] as its input and will have a throws PatternSyntaxException in its declaration?

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
aleitamar: 1st comment:
the default configuration test that the initial value is an empty list.
i think it is necessary, to ensure that by default, no tags will get filtered.

2nd comment:
does it hurt that the test have a throws clause?
1) this is the default eclipse test template
2) if any of the executed code changes to throw a checked exception, the test will not have to change.

i don't mind removing it, but i think it is better to leave it there... (for any test...)

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
aleitamar: this test only checks that the configuration can store multiple patterns...
it is irrelevant what is the actual patterns in this case, as we do not check the actual filtering...

i am not sure i understand what is your suggestion to make it more readable?

src/main/java/net/masterthought/cucumber/Configuration.java
damianszczepanik: 1. OK

2. if regex fails then rethrows https://github.com/damianszczepanik/cucumber-reporting/blob/master/src/main/java/net/masterthought/cucumber/ValidationException.java and that's it. I don't want to have to many methods in configuration

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
aleitamar: :+1:<a href='#crh-update-none'></a>

src/main/java/net/masterthought/cucumber/Configuration.java
aleitamar: :+1:<a href='#crh-update-none'></a>

src/main/java/net/masterthought/cucumber/generators/TagsOverviewPage.java
aleitamar: :+1:<a href='#crh-update-none'></a>

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
aleitamar: :+1:<a href='#crh-update-none'></a>

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
aleitamar: :+1:<a href='#crh-update-complete'></a>

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
aleitamar: :+1:<a href='#crh-update-complete'></a>

aleitamar:
:+1:<a href='#crh-update-none'></a>

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
```diff
@@ -140,4 +146,41 @@ public void getProjectName_ReturnsProjectName() {
...
+        
+        // when
+        Collection<Pattern> patterns = configuration.getTagsToExcludeFromChart();
+
+        // then
+        assertThat(patterns).extractingResultOf("pattern").containsOnly(somePattern, anotherPattern);
+	}
+    
+    @Test
+	public void setTagsToExcludeFromChart_invalidRegexPattern_ThrowsValidationException() throws Exception {
```
aleitamar: As far as i am concerned this test is fine. It fails if the exception is not thrown, and passes when it is thrown.
The static code analysis tool flags this but it is a false positive...

A similar test can be found at:
ReportParserTest.parseJsonResultsFails_OnNoExistingFile_ThrowsException

aleitamar: Of course, if you have any other preferred way of dealing with exceptions in tests, i am open to suggestions...

src/test/java/net/masterthought/cucumber/generators/TagsOverviewPageTest.java
```diff
@@ -1,8 +1,10 @@
 package net.masterthought.cucumber.generators;
 
+import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.util.List;
+import java.util.regex.Pattern;
```
aleitamar: :+1:<a href='#crh-update-none'></a>

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
aleitamar: :+1:<a href='#crh-update-complete'></a>

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
damianszczepanik: 1. comments was only for the method name, not the test itself - keep test if you think it is worth!
2. It does not but it increases noise

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
damianszczepanik: 1. rename to setTagsToExcludeFromChart_OnInvalidRegexPattern_ThrowsValidationException
2. Are you throwing Exception or rather ValidationException?

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
```diff
@@ -140,4 +146,41 @@ public void getProjectName_ReturnsProjectName() {
...
+        // give
+        Configuration configuration = new Configuration(outputDirectory, projectName);
+        thrown.expect(ValidationException.class);
+        
+        // when + then
+        configuration.setTagsToExcludeFromChart("\\invalid.regex\\");
+	}
 }
```
damianszczepanik: btw, add empty line at the end

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
aleitamar: 1. will do
2. throwing ValidationException that wraps the PatternSyntaxException

aleitamar: :+1:<a href='#crh-update-none'></a>

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
aleitamar: :+1:<a href='#crh-update-none'></a>

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
aleitamar: :+1:<a href='#crh-update-none'></a>

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
aleitamar: :+1:<a href='#crh-update-complete'></a>

src/test/java/net/masterthought/cucumber/generators/TagsOverviewPageTest.java
aleitamar: :+1:

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
```diff
@@ -140,4 +146,41 @@ public void getProjectName_ReturnsProjectName() {
...
+        configuration.setTagsToExcludeFromChart(somePattern, anotherPattern);
+
+        // when
+        Collection<Pattern> patterns = configuration.getTagsToExcludeFromChart();
+
+        // then
+        assertThat(patterns).extractingResultOf("pattern").containsOnly(somePattern, anotherPattern);
+    }
+
+    @Test
```
aleitamar: Not a real issue...

aleitamar: :+1:

src/main/java/net/masterthought/cucumber/Configuration.java
```diff
@@ -21,6 +25,8 @@
     private String buildNumber;
     private String projectName;
 
+	private Collection<Pattern> tagsToExcludeFromChart = new ArrayList<>();
```
damianszczepanik: something wrong with indention (formatting)

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
```diff
@@ -140,4 +146,41 @@ public void getProjectName_ReturnsProjectName() {
...
+
+        // when
+        Collection<Pattern> patterns = configuration.getTagsToExcludeFromChart();
+
+        // then
+        assertThat(patterns).isEmpty();
+    }
+
+    @Test
+    public void getTagsToExcludeFromChart_addPatterns_ReturnsListWithAllPatterns() {
```
damianszczepanik: addPatterns -> OnPattern

aleitamar: Since the commit is already merged (thanks :dancer: )
do you still want me to make this change?

damianszczepanik: With next PR :)

Issue #556, opened 2016-11-04
Error showing DocStrings containing html character


fmarmar:
Hi, 

If a DocString content is an xml, reporter fail to show the DocString value.

damianszczepanik:
Provide json report so I can take a look

fmarmar:
Sorry, here it goes. I just made a pull request to fix the issue #557 


[sample-xml.txt](https://github.com/damianszczepanik/cucumber-reporting/files/572741/sample-xml.txt)

mpkorstanje:
This has been fixed in version 3.4.0 with #576 and #584. I think we can close this.

PR #557, opened 2016-11-04
Fix issue #556 

fmarmar:
Set StringEscapeUtils.ESCAPE_HTML4 as escHtml variable in context to use it in docstring.vm to escape docstrings

### DIFF
--- a/src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
+++ b/src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
@@ -8,6 +8,7 @@ import java.io.Writer;
 import java.nio.charset.StandardCharsets;
 import java.util.Properties;
 
+import org.apache.commons.lang3.StringEscapeUtils;
 import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -62,6 +63,7 @@ public abstract class AbstractPage {
 
     private void generateReport() {
         context.put("report_file", getWebPage());
+        context.put("escHtml", StringEscapeUtils.ESCAPE_HTML4);
 
         Template template = engine.getTemplate("templates/generators/" + templateFileName);
         File reportFile = new File(configuration.getReportDirectory(),
--- a/src/main/resources/templates/macros/report/docstring.vm
+++ b/src/main/resources/templates/macros/report/docstring.vm
@@ -8,7 +8,7 @@
           <a>Doc string</a>
       </div>
       <div id="msg-$docStringId" class="collapse collapsable-details">
-        <pre>$docString.getValue()</pre>
+        <pre>$escHtml.translate($docString.getValue())</pre>
       </div>
     </div>
   </div>
--- a/src/test/java/net/masterthought/cucumber/ReportGenerator.java
+++ b/src/test/java/net/masterthought/cucumber/ReportGenerator.java
@@ -18,6 +18,7 @@ public abstract class ReportGenerator {
     public final static String JSON_DIRECTORY = "json/";
 
     protected static final String SAMPLE_JSON = "sample.json";
+    protected static final String SAMPLE_XML_JSON = "sample-xml.json";
     public static final String SIMPLE_JSON = "simple.json";
     protected static final String EMPTY_JSON = "empty.json";
     protected static final String INVALID_JSON = "invalid.json";
--- a/src/test/java/net/masterthought/cucumber/generators/integrations/FeatureReportPageIntegrationTest.java
+++ b/src/test/java/net/masterthought/cucumber/generators/integrations/FeatureReportPageIntegrationTest.java
@@ -281,6 +281,27 @@ public class FeatureReportPageIntegrationTest extends PageTest {
 
         stepElement.getDocString().hasDocString(step.getDocString());
     }
+    
+    @Test
+    public void generatePage_generatesEscapedDocString() {
+
+        // given
+        setUpWithJson(SAMPLE_XML_JSON);
+        final Feature feature = features.get(0);
+        page = new FeatureReportPage(reportResult, configuration, feature);
+
+        // when
+        page.generatePage();
+
+        // then
+        DocumentAssertion document = documentFrom(page.getWebPage());
+
+        StepAssertion stepElement = document.getFeature().getElements()[0].getStepsSection().getSteps()[3];
+        Step step = feature.getElements()[0].getSteps()[3];
+
+        stepElement.getDocString().hasDocString(step.getDocString());
+    }
+
 
     @Test
     public void generatePage_generatesEmbedding() {
--- a/src/test/java/net/masterthought/cucumber/generators/integrations/helpers/DocStringAssertion.java
+++ b/src/test/java/net/masterthought/cucumber/generators/integrations/helpers/DocStringAssertion.java
@@ -1,13 +1,12 @@
 package net.masterthought.cucumber.generators.integrations.helpers;
 
-import org.apache.commons.lang3.StringEscapeUtils;
+import static org.assertj.core.api.Assertions.assertThat;
+import net.masterthought.cucumber.json.DocString;
+
 import org.apache.commons.lang3.text.translate.AggregateTranslator;
 import org.apache.commons.lang3.text.translate.CharSequenceTranslator;
 import org.apache.commons.lang3.text.translate.EntityArrays;
 import org.apache.commons.lang3.text.translate.LookupTranslator;
-
-import net.masterthought.cucumber.json.DocString;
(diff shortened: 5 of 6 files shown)


src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -62,6 +63,7 @@ public void generatePage() {
 
     private void generateReport() {
         context.put("report_file", getWebPage());
+        context.put("escHtml", StringEscapeUtils.ESCAPE_HTML4);
```
damianszczepanik: You are assuming that every doc uses HTML4 what might not be true

src/test/java/net/masterthought/cucumber/ReportGenerator.java
```diff
@@ -18,6 +18,7 @@
     public final static String JSON_DIRECTORY = "json/";
 
     protected static final String SAMPLE_JSON = "sample.json";
+    protected static final String SAMPLE_XML_JSON = "sample-xml.json";
```
damianszczepanik: you don't need to add json to test it, you can test the code as unit test rather than integration

src/main/resources/templates/macros/report/docstring.vm
```diff
@@ -8,7 +8,7 @@
           <a>Doc string</a>
       </div>
       <div id="msg-$docStringId" class="collapse collapsable-details">
-        <pre>$docString.getValue()</pre>
+        <pre>$escHtml.translate($docString.getValue())</pre>
```
fmarmar: I don't understand. If you write a gherkin scenario and you put an html in the doc string, i assume is what you want to see in the report. For example:

Given the next html
'''
<b>Bold Text</b>
'''
When I open it in a browser
Then I should see the text "Bold Text" render as bold

When I see the report, I would like to see the raw text

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
fmarmar: No, I just setting a variable that allows me to use a function to escape characters into html entities in any velocity temple.

src/test/java/net/masterthought/cucumber/ReportGenerator.java
fmarmar: Sorry, given my solution is to modify the velocity template I don't see how I can test without an integration test.

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
damianszczepanik: but this is HTML4 implementation and other may not work

src/main/resources/templates/macros/report/docstring.vm
damianszczepanik: Not really. Developer can use this option to display any HTML content so they can control how the browser renders it (eg display results in HTML table). If they want to see HTML code as escaped content they format that content before storing into the report.

By this change you don't allow developers to display additional information as HTML table, formatted text or image loaded from external server

src/test/java/net/masterthought/cucumber/ReportGenerator.java
damianszczepanik: unit test?

src/main/resources/templates/macros/report/docstring.vm
fmarmar: If a developer is writing a bdd specification thinking how it will render in an html report, in my opinion is a bad specification. Anyway, what do you propose? We can set a flag in the configuration object to escape the docStrings.

damianszczepanik:
Maybe I don't understand the idea o docstring and the reason why it should be escaped instead of printed as it is? By escaping you loose formatting

mpkorstanje:
If I understand the motivation correctly:

Feature files are plain text and contain no markup besides whitespace formatting. The reports should accurately represent these files. So when printed they should escape any html to avoid losing that information. In this context the HTML is not formatting.

So when a feature file contains:

Given some text with html
''''
<bold>a bold statement</bold>
"""

This  be rendered as 

Given some text with html
''''
<bold>a bold statement</bold>
"""

Not:

Given some text with html
''''
**a bold statement**
"""

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
mpkorstanje: From the documentation of [StringEscapeUtils.html#escapeHtml4](https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/StringEscapeUtils.html#escapeHtml4(java.lang.String)): 
_Supports all known HTML 4.0 entities, including funky accents. Note that the commonly used apostrophe escape character (&apos;) is not a legal entity and so is not supported)._ 

And from the [Wikipedia entry on HTML entities](https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references#Character_entity_references_in_HTML) we can see that none have been added for HTML5. Footnotes q and r are interesting but shouldn't be an issue for any modern browser.

damianszczepanik:
Ok you attach problematic JSON file?

mpkorstanje:
Feature:

```cucumber
Feature: <b>a bold statement</b>

  <b>a bold statement</b>

  Scenario: <b>a bold statement</b>
  <b>a bold statement</b>

    Given some <b>text</b> with html
    """
    <b>a bold
    statement</b>
    """

    Given some table with html
      | <b>bold</b>             | <i>beautiful</i> |

```

Results in json:

```json
[
  {
    "line": 1,
    "elements": [
      {
        "line": 5,
        "name": "\u003cb\u003ea bold statement\u003c/b\u003e",
        "description": "\u003cb\u003ea bold statement\u003c/b\u003e",
        "id": "\u003cb\u003ea-bold-statement\u003c/b\u003e;\u003cb\u003ea-bold-statement\u003c/b\u003e",
        "after": [
          {
            "result": {
              "duration": 74315,
              "status": "passed"
            },
            "match": {
              "location": "LambdaGlueBase.java:56"
            }
          }
        ],
        "type": "scenario",
        "keyword": "Scenario",
        "steps": [
          {
            "result": {
              "duration": 60167854,
              "status": "passed"
            },
            "line": 8,
            "name": "some \u003cb\u003etext\u003c/b\u003e with html",
            "match": {
              "location": "MyStepdefs.java:20"
            },
            "keyword": "Given ",
            "doc_string": {
              "content_type": "",
              "line": 9,
              "value": "\u003cb\u003ea bold\nstatement\u003c/b\u003e"
            }
          },
          {
            "result": {
              "duration": 667535,
              "status": "passed"
            },
            "line": 14,
            "name": "some table with html",
            "match": {
              "location": "MyStepdefs.java:17"
            },
            "rows": [
              {
                "cells": [
                  "\u003cb\u003ebold\u003c/b\u003e",
                  "\u003ci\u003ebeautiful\u003c/i\u003e"
                ],
                "line": 15
              }
            ],
            "keyword": "Given "
          }
        ]
      }
    ],
    "name": "\u003cb\u003ea bold statement\u003c/b\u003e",
    "description": "\n\u003cb\u003ea bold statement\u003c/b\u003e",
    "id": "\u003cb\u003ea-bold-statement\u003c/b\u003e",
    "keyword": "Feature",
    "uri": "com/github/mpkorstanje/cucumber/cukes.feature"
  }
]

```


Renders in the console as:

```
[INFO] --- maven-surefire-plugin:2.19.1:test (default-test) @ cucumber ---

-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running com.github.mpkorstanje.cucumber.RunCukesTest
Feature: <b>a bold statement</b>
  
  <b>a bold statement</b>

  Scenario: <b>a bold statement</b>  # com/github/mpkorstanje/cucumber/cukes.feature:5
    <b>a bold statement</b>

    Given some <b>text</b> with html # MyStepdefs.java:20
      """
      <b>a bold
      statement</b>
      """
    Given some table with html       # MyStepdefs.java:17

1 Scenarios (1 passed)
2 Steps (2 passed)
0m0.063s

Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.565 sec - in com.github.mpkorstanje.cucumber.RunCukesTest
```

But renders in the browser as

screenshot from 2016-12-12 22-58-28

mpkorstanje:
This one is particularly nasty.

```cucumber
Feature: XXS
  Scenario: Show an alert
    Given some <b>text</b> with html
    """
    </pre><script>alert('XXS!');</script><pre>
    """

```

mpkorstanje:
If neither  @fmarmar nor @damianszczepanik can solve these I'll try to send a pull request sometime next week.

damianszczepanik:
Supported in pull #580

PR #561, opened 2016-11-16
Move scenario column to be in same order as trends

stevejensen22:
In my experience, stakeholders are confused by the Steps section, and really only care about Scenarios. This pull request proposes to swap the postion of the Steps and Scenarios sections to alleviate this confusion. It also puts the organization in congruence with the Trends pages.

screen shot 2016-11-15 at 10 32 08 pm

screen shot 2016-11-15 at 10 30 35 pm

### DIFF
--- a/src/main/resources/templates/macros/report/reportHeader.vm
+++ b/src/main/resources/templates/macros/report/reportHeader.vm
@@ -6,8 +6,8 @@
       <th></th>
     #end
     <th></th>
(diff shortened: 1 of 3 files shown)


damianszczepanik:
I understand your point of view,  my intention was to have features that contain scenarios that contain steps. The order is not very important as those relations are unchanged.

PR #576, opened 2016-12-14
 Escape HTML using the velocity engine

mpkorstanje:
It is preferable to escape HTML through the template engine over escaping it at in
the template. This ensures none can be forgotten on accident. The unfortunate side
effect of this is that it no longer possible to  provide html as an argument in a
macro. This leads to some duplication in embeddings.vm

 Existing methods to escape html have been removed.

### DIFF
--- a/src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
+++ b/src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
@@ -14,6 +14,7 @@ import org.apache.logging.log4j.Logger;
 import org.apache.velocity.Template;
 import org.apache.velocity.VelocityContext;
 import org.apache.velocity.app.VelocityEngine;
+import org.apache.velocity.app.event.EventCartridge;
 import org.apache.velocity.runtime.resource.loader.ClasspathResourceLoader;
 
 import net.masterthought.cucumber.Configuration;
@@ -84,6 +85,11 @@ public abstract class AbstractPage {
     }
 
     private void buildGeneralParameters() {
+        // to escape html and xml
+        EventCartridge ec = new EventCartridge();
+        ec.addEventHandler(new EscapeHtmlReference());
+        context.attachEventCartridge(ec);
+
         // to provide unique ids for elements on each page
         context.put("counter", new Counter());
         context.put("util", Util.INSTANCE);
--- a/src/main/java/net/masterthought/cucumber/json/Result.java
+++ b/src/main/java/net/masterthought/cucumber/json/Result.java
@@ -1,8 +1,6 @@
 package net.masterthought.cucumber.json;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
-import org.apache.commons.lang.StringEscapeUtils;
-
 import net.masterthought.cucumber.json.support.Status;
 import net.masterthought.cucumber.util.Util;
 
@@ -27,8 +25,7 @@ public class Result {
         return Util.formatDuration(duration);
     }
 
-    /** Returns error message escaped so it is properly displayed as HTML. */
     public String getErrorMessage() {
-        return StringEscapeUtils.escapeHtml(errorMessage);
+        return errorMessage;
     }
 }
--- a/src/main/resources/templates/macros/json/embeddings.vm
+++ b/src/main/resources/templates/macros/json/embeddings.vm
@@ -1,8 +1,8 @@
 #macro(includeEmbeddings, $embeddings)
 
-#if (!$embeddings.isEmpty())
+  #if (!$embeddings.isEmpty())
   <div class="embeddings inner-level">
-  #foreach($embedding in $embeddings)
+    #foreach($embedding in $embeddings)
     #if ($embedding.getMimeType() == "image/png")
       #includeImageEmbedding($embedding, "png", $foreach.index)
     #elseif ($embedding.getMimeType() == "image/bmp")
@@ -10,7 +10,7 @@
     #elseif ($embedding.getMimeType() == "image/jpeg")
       #includeImageEmbedding($embedding, "jpeg", $foreach.index)
     #elseif ($embedding.getMimeType() == "text/xml")
-      #includeEscapedTextEmbedding($embedding, "XML text", $foreach.index)
+      #includeTextEmbedding($embedding, "XML text", $foreach.index)
     #elseif ($embedding.getMimeType() == "text/plain")
       #includeTextEmbedding($embedding, "Plain text", $foreach.index)
     #elseif ($embedding.getMimeType() == "text/html")
@@ -24,46 +24,78 @@
     #end
   #end
   </div>
-#end
+  #end
 
 #end
 
 
 
 #macro(includeImageEmbedding, $embedding, $image_type, $index)
-  #set($Quote = '"')
-  #includeExpandableEmbedding($embedding, $image_type, "<img src=${Quote}embeddings/$embedding.getFileName()${Quote}>", $index)
-#end
-
-#macro(includeImageReferenceEmbedding, $embedding, $index)
-  #set($Quote = '"')
-  #includeExpandableEmbedding($embedding, "Image", "<img src=${Quote}$embedding.getDecodedData()${Quote}>", $index)
-#end
-
-#macro(includeTextEmbedding, $embedding, $format, $index)
-  #includeExpandableEmbedding($embedding, $format, "<pre>$embedding.getDecodedData()</pre>", $index)
-#end
-
-#macro(includeEscapedTextEmbedding, $embedding, $format, $index)
-  #includeExpandableEmbedding($embedding, $format, "<pre>$embedding.getEscapedDecodedData()</pre>", $index)
-#end
-
-#macro(includeUnknownEmbedding, $embedding, $index)
-  #set($Quote = '"')
-  #includeExpandableEmbedding($embedding, $embedding.getMimeType(), "This file cannot be displayed. Use download button to get the content as file.", $index)
-#end
-
-
-#macro(includeExpandableEmbedding, $embedding, $type, $content, $index)
   #set($index = $index + 1)
   <div class="embedding indention">
     #set($embeddingId = $counter.next())
     <div data-toggle="collapse" data-target="#embedding-$embeddingId" class="collapsable-control">
-      <a>Attachment $index ($type)</a>
-      <a href="embeddings/$embedding.getFileName()" target="_blank"><span class="download-button glyphicon glyphicon-download-alt"></span></a>
+      <a>Attachment $index ($image_type)</a>
+      <a href="embeddings/$embedding.getFileName()" target="_blank">
+        <span class="download-button glyphicon glyphicon-download-alt"></span>
+      </a>
     </div>
     <div id="embedding-$embeddingId" class="collapse collapsable-details">
-      <span class="embedding-content">$content</span>
+      <span class="embedding-content">
+        <img src="embeddings/$embedding.getFileName()">
+      </span>
+    </div>
+  </div>
+#end
+
+#macro(includeImageReferenceEmbedding, $embedding, $index)
+  #set($index = $index + 1)
+  <div class="embedding indention">
+    #set($embeddingId = $counter.next())
+    <div data-toggle="collapse" data-target="#embedding-$embeddingId" class="collapsable-control">
+      <a>Attachment $index (Image)</a>
+      <a href="embeddings/$embedding.getFileName()" target="_blank">
+        <span class="download-button glyphicon glyphicon-download-alt"></span>
+      </a>
+    </div>
+    <div id="embedding-$embeddingId" class="collapse collapsable-details">
+      1<span class="embedding-content">
+        <img src="$embedding.getDecodedData()">
+      </span>
+    </div>
+  </div>
+#end
+
+#macro(includeTextEmbedding, $embedding, $format, $index)
+  #set($index = $index + 1)
+  <div class="embedding indention">
+    #set($embeddingId = $counter.next())
+    <div data-toggle="collapse" data-target="#embedding-$embeddingId" class="collapsable-control">
+      <a>Attachment $index ($format)</a>
+      <a href="embeddings/$embedding.getFileName()" target="_blank">
+        <span class="download-button glyphicon glyphicon-download-alt"></span>
+      </a>
+    </div>
+    <div id="embedding-$embeddingId" class="collapse collapsable-details">
+      <span class="embedding-content">
+        <pre>$embedding.getDecodedData()</pre>
+      </span>
+    </div>
+  </div>
+#end
+
+#macro(includeUnknownEmbedding, $embedding, $index)
+  #set($index = $index + 1)
+  <div class="embedding indention">
+    #set($embeddingId = $counter.next())
+    <div data-toggle="collapse" data-target="#embedding-$embeddingId" class="collapsable-control">
+      <a>Attachment $index ($embedding.getMimeType())</a>
+      <a href="embeddings/$embedding.getFileName()" target="_blank">
+        <span class="download-button glyphicon glyphicon-download-alt"></span>
+      </a>
+    </div>
+    <div id="embedding-$embeddingId" class="collapse collapsable-details">
+        <span class="embedding-content">This file cannot be displayed. Use download button to get the content as file.</span>
     </div>
   </div>
 #end
--- a/src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
+++ b/src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
@@ -17,6 +17,7 @@ import net.masterthought.cucumber.ReportBuilder;
 import net.masterthought.cucumber.Trends;
 import net.masterthought.cucumber.ValidationException;
 import net.masterthought.cucumber.generators.integrations.PageTest;
+import net.masterthought.cucumber.generators.integrations.helpers.DocumentAssertion;
 import net.masterthought.cucumber.util.Counter;
 import net.masterthought.cucumber.util.Util;
 
@@ -46,6 +47,33 @@ public class AbstractPageTest extends PageTest {
         File reportFile = new File(configuration.getReportDirectory(),
                 ReportBuilder.BASE_DIRECTORY + File.separatorChar + page.getWebPage());
         assertThat(reportFile).exists();
+    }
+
+
+    @Test
+    public void generateReport_DisplaysContentAsEscapedText() {
+
+        // given
+        page = new FeatureReportPage(reportResult, configuration, features.get(1));
+
+        // when
+        page.generatePage();
+
+        // then
+        DocumentAssertion document = documentFrom(page.getWebPage());
+        assertThat(document.getFeature().getDescription())
+                .isEqualTo("As an Account Holder I want to withdraw cash from an ATM,<br>so that I can get money when the bank is closed");
+        assertThat(document.getFeature().getElements()[0].getStepsSection().getSteps()[5].getEmbedding()[3].text())
+                .isEqualTo("Attachment 4 (HTML) <i>Hello</i> <b>World!</b>");
+        assertThat(document.getFeature().getElements()[0].getStepsSection().getSteps()[5].getMessage().text())
+                .isEqualTo("Error message java.lang.AssertionError: \n" +
+                        "Expected: is <80>\n" +
+                        "     got: <90>\n" +
+                        "\n" +
+                        "\tat org.junit.Assert.assertThat(Assert.java:780)\n" +
+                        "\tat org.junit.Assert.assertThat(Assert.java:738)\n" +
+                        "\tat net.masterthought.example.ATMScenario.checkBalance(ATMScenario.java:69)\n" +
+                        "\tat ✽.And the account balance should be 90(net/masterthought/example/ATMK.feature:12)");
     }
 
     @Test
--- a/src/test/java/net/masterthought/cucumber/generators/EscapeHtmlReferenceTest.java
+++ b/src/test/java/net/masterthought/cucumber/generators/EscapeHtmlReferenceTest.java
@@ -1,0 +1,34 @@
+package net.masterthought.cucumber.generators;
+
+import org.apache.velocity.app.event.ReferenceInsertionEventHandler;
+import org.junit.Test;
+
+import static org.apache.commons.lang.StringEscapeUtils.escapeHtml;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author M.P. Korstanje (mpkorstanje@github)
+ */
+public class EscapeHtmlReferenceTest {
+
+    private static final String SOME_REFERENCE = "someReference";
+    private final ReferenceInsertionEventHandler insertionEventHandler = new EscapeHtmlReference();
+
+    @Test
+    public void referenceInsert_returnNormalText(){
+        String normalText = "a plain statement";
+        assertThat(insertionEventHandler.referenceInsert(SOME_REFERENCE, normalText)).isEqualTo(normalText);
+    }
+
+    @Test
+    public void referenceInsert_shouldEscapeHtmlForAnyLabel(){
+        String html = "<b>a bold statement</b>";
+        assertThat(insertionEventHandler.referenceInsert(SOME_REFERENCE, html)).isEqualTo(escapeHtml(html));
+    }
+
+    @Test
+    public void referenceInsert_shouldReturnNullForNull(){
+        assertThat(insertionEventHandler.referenceInsert(SOME_REFERENCE, null)).isNull();
+    }
+
+}
--- a/src/test/java/net/masterthought/cucumber/generators/integrations/FailuresOverviewPageIntegrationTest.java
+++ b/src/test/java/net/masterthought/cucumber/generators/integrations/FailuresOverviewPageIntegrationTest.java
@@ -84,6 +84,6 @@ public class FailuresOverviewPageIntegrationTest extends PageTest {
         DocumentAssertion document = documentFrom(page.getWebPage());
         ElementAssertion[] elements = document.getElements();
         assertThat(elements).hasSize(1);
-        assertThat(elements[0].getBrief().getName()).isEqualTo(features.get(1).getElements()[0].getEscapedName());
+        assertThat(elements[0].getBrief().getName()).isEqualTo(features.get(1).getElements()[0].getName());
     }
 }
--- a/src/test/java/net/masterthought/cucumber/generators/integrations/helpers/EmbeddingAssertion.java
+++ b/src/test/java/net/masterthought/cucumber/generators/integrations/helpers/EmbeddingAssertion.java
@@ -19,7 +19,8 @@ public class EmbeddingAssertion extends ReportAssertion {
     }
 
     public void hasTextContent(String content) {
-        assertThat(getBox().oneBySelector("pre", WebAssertion.class).html()).isEqualTo(getDecodedData(content));
+        assertThat(getBox().oneBySelector("pre", WebAssertion.class).text())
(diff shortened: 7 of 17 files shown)


mpkorstanje:
First draft of a fundamental solution for  #556. 

Still a WIP but open for comments.

mpkorstanje:
Visual inspection of sample.json report looks good.

pom.xml
```diff
@@ -228,6 +228,13 @@
             <artifactId>velocity-tools</artifactId>
             <version>1.4</version>
         </dependency>
+
+        <dependency>
+            <groupId>oro</groupId>
+            <artifactId>oro</artifactId>
+            <version>2.0.8</version>
+        </dependency>
```
damianszczepanik: I was not able to run this code because of
``
ERROR StatusLogger No log4j2 configuration file found. Using default configuration: logging only errors to the console.
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/oro/text/perl/MalformedPerl5PatternException
``

src/main/resources/templates/macros/json/embeddings.vm
```diff
@@ -31,39 +31,63 @@
 
 
 #macro(includeImageEmbedding, $embedding, $image_type, $index)
-  #set($Quote = '"')
-  #includeExpandableEmbedding($embedding, $image_type, "<img src=${Quote}embeddings/$embedding.getFileName()${Quote}>", $index)
+  #set($index = $index + 1)
+  <div class="embedding indention">
+    #set($embeddingId = $counter.next())
```
damianszczepanik: what is going on here, why the change is so big?

src/test/java/net/masterthought/cucumber/generators/integrations/FailuresOverviewPageIntegrationTest.java
```diff
@@ -69,21 +69,4 @@ public void generatePage_onJsonWithoutFailedSteps_generatesProperMessage() {
         SummaryAssertion summary = document.getReport();
         assertThat(summary.getEmptyReportMessage()).isEqualTo("You have no failed scenarios in your Cucumber report");
     }
```
damianszczepanik: why those tests were removed

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -79,6 +80,7 @@ private Properties buildProperties() {
         props.setProperty("resource.loader", "class");
         props.setProperty("class.resource.loader.class", ClasspathResourceLoader.class.getCanonicalName());
         props.setProperty("runtime.log", new File(configuration.getReportDirectory(), "velocity.log").getPath());
+        props.setProperty("eventhandler.referenceinsertion.class", "org.apache.velocity.app.event.implement.EscapeHtmlReference");
```
damianszczepanik: so by this all labels are escaped so you are not able to provide HTML code into the report and with it will be displayed as HTML (eg link to external server in description)?

pom.xml
mpkorstanje: You may need to do a `mvn clean install` first or if you're using an IDE reload or refresh the project. Looks like it is not downloading the dependency.

damianszczepanik: mvn clean test works fine so dependencies are fine

src/test/java/net/masterthought/cucumber/generators/integrations/FailuresOverviewPageIntegrationTest.java
mpkorstanje: They were testing methods that escaped data. These methods were no longer used and thus removed, leaving nothing to be tested.

src/main/resources/templates/macros/json/embeddings.vm
mpkorstanje: The `includeExpandableEmbedding` macro was used like this:

    #includeExpandableEmbedding($embedding, $format,
                <pre>$embedding.getEscapedDecodedData()</pre>", $index) 

Escaping all html broke the  macro because the value of `$content` would also be escaped. I fixed this by inlining `includeExpandableEmbedding` into every macro that used it. This results in the heavy duplication that you're looking at now.

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
mpkorstanje: The way it is currently setup that is correct. 

However it is possible to exclude some labels from being escaped. Though I would not recommend this. Rather I would suggest adding a markdown parser and use markdown formatting to provide the markup.

You can read about the functions and options that velocity provides here:

https://velocity.apache.org/engine/devel/developer-guide.html#orgapachevelocityappeventreferenceinsertioneventhandler

pom.xml
mpkorstanje: Interesting. You should not be able to pass the tests without oro. EscapeReference the parent of EscapeHtmlReference uses oro. Can you provide a simple reproducer?

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
mpkorstanje: Having looked at [flexmark](https://github.com/vsch/flexmark-java)  I think it would be fairly simple to add. Though I also think that should be done on another pull request.

mpkorstanje: But I suppose that allowing some html elements in the feature section is permissible.

```
<a>              - hyperlink.
<b>              - bold, use as last resort <h1>-<h3>, <em>, and <strong> are 
                   preferred.
<blockquote>     - specifies a section that is quoted from another source.
<code>           - defines a piece of computer code.
<del>            - delete, used to indicate modifications.
<dd>             - describes the item in a <dl> description list.
<dl>             - description list.
<dt>             - title of an item in a <dl> description list.
<em>             - emphasized.
<h1>, <h2>, <h3> - headings.
<i>              - italic.
<img>            - specifies an image tag.
<kbd>            - represents user input (usually keyboard input).
<li>             - list item in an ordered list <ol> or an unordered list <ul>.
<ol>             - ordered list.
<p>              - paragraph.
<pre>            - pre-element displayed in a fixed width font and and 
                   unchanged line breaks.
<s>              - strikethrough.
<sup>            - superscript text appears 1/2 character above the baseline 
                   used for footnotes and other formatting.
<sub>            - subscript appears 1/2 character below the baseline.
<strong>         - defines important text.
<strike>         - strikethrough is deprecated, use <del> instead.
<ul>             - unordered list.
<br>             - line break.
<hr>             - defines a thematic change in the content, usually via a 
                   horizontal line.
```

pom.xml
damianszczepanik: https://github.com/damianszczepanik/cucumber-reporting#usage here you have some code how to run the report as the standalone app - I use this code and then the code fails.

You can use also this real code from the Jenkins plugin https://github.com/jenkinsci/cucumber-reports-plugin/blob/master/src/main/java/net/masterthought/jenkins/CucumberReportPublisher.java

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
damianszczepanik: I'm not sure If I want to display doc_string should be displayed as escaped html but let's go for this and if there will be other requests I can handle this

src/test/java/net/masterthought/cucumber/generators/integrations/FailuresOverviewPageIntegrationTest.java
damianszczepanik: But then we have no test that validates if data is escaped properly (which is not global. default configuration). Worth to have test for this, even one.

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
damianszczepanik: Also I prefer compile time checking, not runtime checking
EscapeHtmlReference.class.getName() should work

mpkorstanje: I'm sorry. I can't quite parse your intentions.

Should the report look like this

> 
> Scenario: A scenario
> 
> <i>Some text here with safe html markup.</i>
> 
> Feature: Docstring
>    Given a doc string containing html mark up
>    """
>   \<i\>Some text here with html markup\</i\>
>    """
> 

Or like this:

> 
> Scenario: A scenario
> 
> \<i\>Some text here with safe html markup.\</i\>
> 
> Feature: Docstring
>    Given a doc string containing html mark up
>    """
>   \<i\>Some text here with html markup\</i\>
>    """
>

src/test/java/net/masterthought/cucumber/generators/integrations/FailuresOverviewPageIntegrationTest.java
damianszczepanik: let's leave separate

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
damianszczepanik: First one as I don't want to change data from the json file. Whatever you pass as doc_string it will be displayed as it is. This allows you to have html code in generated report. If you want to escape html into plain text maybe you should do it by yourself before report is started?

I have doubts for this so at this point all cases are fine as long as more users expect to have doc_string escaped let's leave version you prepared. In case of problem we will investigate and prepare new PR

mpkorstanje: Mmh. I don't understand why you call it 'unchanged'

Suppose I have a json file that contains:

    <b> a bold statement </b>

When rendered by a plain text reporter, like the command-line, it will be rendered as 

    <b> a bold statement </b>

And when rendered by a html reporter that doesn't escape the input
<b> a bold statement </b>
Because now I've lost the \<b\> \</b\> tags from the report. 

---

I also don't understand why you expect the html to be escaped in the json file if it should be displayed escaped. When creating the json file and escaping html I've coupled that json file to a particular report formatter (one that uses html). This seems bad practice.

pom.xml
```diff
@@ -228,6 +228,7 @@
             <artifactId>velocity-tools</artifactId>
             <version>1.4</version>
         </dependency>
+
```
damianszczepanik: revert the whole file if nothing has changed

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -84,6 +88,11 @@ private Properties buildProperties() {
     }
 
     private void buildGeneralParameters() {
+        // to escape html
```
damianszczepanik: html and xml :)

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -107,4 +116,20 @@ private void buildGeneralParameters() {
...
     }
+
+    /**
+     * Escapes all html that was provided in a reference before inserting it into a template.
+     */
+    private final static class EscapeHtmlReference implements ReferenceInsertionEventHandler {
+
+        @Override
+        public Object referenceInsert(String reference, Object value) {
+            if (value == null) {
```
damianszczepanik: when the value can be null?

src/main/java/net/masterthought/cucumber/json/Result.java
```diff
@@ -27,8 +25,7 @@ public String getFormatedDuration() {
         return Util.formatDuration(duration);
     }
 
-    /** Returns error message escaped so it is properly displayed as HTML. */
     public String getErrorMessage() {
-        return StringEscapeUtils.escapeHtml(errorMessage);
+        return errorMessage;
```
damianszczepanik: did you remove 'escape' operation from all methods?

src/main/resources/templates/macros/json/embeddings.vm
```diff
@@ -10,7 +10,7 @@
     #elseif ($embedding.getMimeType() == "image/jpeg")
       #includeImageEmbedding($embedding, "jpeg", $foreach.index)
     #elseif ($embedding.getMimeType() == "text/xml")
-      #includeEscapedTextEmbedding($embedding, "XML text", $foreach.index)
+      #includeTextEmbedding($embedding, "XML text", $foreach.index)
```
damianszczepanik: just to be sure 'includeEscapedTextEmbedding' was removed from the code?

src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
```diff
@@ -48,8 +47,32 @@ public void generateReport_CreatesReportFile() {
         assertThat(reportFile).exists();
     }
 
+
+    @Test
+    public void generateReport_EscapesHtml() {
```
damianszczepanik: generateReport_DisplaysContentAsEscapedText

src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
```diff
@@ -48,8 +47,32 @@ public void generateReport_CreatesReportFile() {
...
+                equalTo("<b>a bold statement</b>"));
+        assertThat(document.getFeature().getElements()[0].getStepsSection().getSteps()[0].getDocString().text(),
+                equalTo("Doc string <b>a bold\nstatement</b>"));
+        assertThat(document.getFeature().getElements()[0].getStepsSection().getSteps()[1].getArgumentsTable().getBodyRow().getCellsValues()[0],
+                equalTo("<b>bold</b>"));
+    }
+
     @Test
     public void generateReport_OnInvalidPath_ThrowsException() {
+        setUpWithJson(SAMPLE_JSON);
```
damianszczepanik: why it was not needed and now you copy it over and over again?

mpkorstanje:
I've removed the dependency on Oro. Does it work for you now?

src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
```diff
@@ -1,11 +1,14 @@
 package net.masterthought.cucumber.generators;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.hamcrest.CoreMatchers.equalTo;
```
damianszczepanik: we don't use this equal method, compare with other tests

src/test/resources/json/html.json
```diff
@@ -0,0 +1,71 @@
+[
```
damianszczepanik: don't need to have separate JSON if you use current sample JSON or test via unit test

mpkorstanje: The current sample json doesn't contain enough html.

I'd also rather not modify it as FeaturesOverviewPageIntegrationTest#generatePage_generatesStatsTableFooter and some other tests rely on statistics of that exact file.

src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
mpkorstanje: Because leaving it in the @Before method and then calling setUpWithJson(HTML_JSON); in the escapeHtml test would be deceptive.

src/main/resources/templates/macros/json/embeddings.vm
mpkorstanje: Yes. There is no need for it any more.

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
mpkorstanje: There is no specification so I copied that detail from the EscapeReference implementation but if you follow the data flow up to ASTReference.render you'll see that null is a plausible value.

Presumably when using

```
 $object.methodThatReturnsNull()
```

src/main/java/net/masterthought/cucumber/json/Result.java
mpkorstanje: Except in a few tests that were now comparing unescaped text against escaped text.

mpkorstanje:
Commits have been squashed and force pushed. They're in this PR now.

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
damianszczepanik: OK

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -107,4 +115,20 @@ private void buildGeneralParameters() {
             }
         }
     }
+
+    /**
+     * Escapes all html and xml that was provided in a reference before inserting it into a template.
+     */
+    private final static class EscapeHtmlReference implements ReferenceInsertionEventHandler {
```
damianszczepanik: As you see there are many tests (not only unit) that reduces number of regressions when preparing new PR.
Can you add unit test for both cases?

src/main/resources/templates/macros/json/embeddings.vm
```diff
@@ -31,39 +31,63 @@
...
-
-#macro(includeExpandableEmbedding, $embedding, $type, $content, $index)
   #set($index = $index + 1)
   <div class="embedding indention">
-    #set($embeddingId = $counter.next())
+  #set($embeddingId = $counter.next())
     <div data-toggle="collapse" data-target="#embedding-$embeddingId" class="collapsable-control">
-      <a>Attachment $index ($type)</a>
-      <a href="embeddings/$embedding.getFileName()" target="_blank"><span class="download-button glyphicon glyphicon-download-alt"></span></a>
+        <a>Attachment $index ($embedding.getMimeType())</a>
```
damianszczepanik: I use two spaces indention, if you don't mind

src/test/java/net/masterthought/cucumber/generators/integrations/FailuresOverviewPageIntegrationTest.java
```diff
@@ -69,21 +69,4 @@ public void generatePage_onJsonWithoutFailedSteps_generatesProperMessage() {
...
-
-    @Test
-    public void generatePage_generatesSummary() {
-
```
damianszczepanik: this test checks if for the 'failure overview' page there is only one scenario with expected name - removing this test we loose validation what is really displayed on the page

src/test/java/net/masterthought/cucumber/json/FeatureTest.java
```diff
@@ -46,20 +46,6 @@ public void getId_ReturnsID() {
     }
 
     @Test
-    public void getElements_ReturnsElements() {
-
-        // given
-        Feature feature = features.get(0);
-
-        // when
-        Element[] elements = feature.getElements();
```
damianszczepanik: do we have unit test for getElements() now?

src/test/resources/json/html.json
damianszczepanik: I know this is not the prefect solution as it is sometimes pain to maintain but as long as there are so many tests - tests says which steps should be updated and it's easy

I prefer to have one, complex json file with all options than having many files. Add required items to existing file (only required items

damianszczepanik: BTW, you can rebase code

App is working correctly now (when you removed lib dependency)

src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
```diff
@@ -48,8 +43,32 @@ public void generateReport_CreatesReportFile() {
...
+
+        // when
+        page.generatePage();
+
+        // then
+        DocumentAssertion document = documentFrom(page.getWebPage());
+        assertThat(document.getFeature().getDescription())
+                .isEqualTo("<b>a bold statement</b>");
+        assertThat(document.getFeature().getElements()[0].getDescription())
+                .isEqualTo("<b>a bold statement</b>");
```
damianszczepanik: name of the method says that data should be escaped while tests validates against not-escaped content - can you tell me more about it?

src/main/resources/templates/macros/json/embeddings.vm
```diff
@@ -31,39 +31,63 @@
...
+        <a href="embeddings/$embedding.getFileName()" target="_blank"><span class="download-button glyphicon glyphicon-download-alt"></span></a>
     </div>
     <div id="embedding-$embeddingId" class="collapse collapsable-details">
-      <span class="embedding-content">$content</span>
+        <span class="embedding-content">This file cannot be displayed. Use download button to get the content as file.</span>
     </div>
   </div>
-#end
+#end
```
damianszczepanik: empty line is missing

src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
```diff
@@ -1,24 +1,23 @@
...
 import net.masterthought.cucumber.util.Counter;
 import net.masterthought.cucumber.util.Util;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.velocity.VelocityContext;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import java.io.File;
+import java.util.Properties;
```
damianszczepanik: usually java package is located the most top, but anyway

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -84,6 +87,11 @@ private Properties buildProperties() {
     }
 
     private void buildGeneralParameters() {
+        // to escape html and xml
+        EventCartridge ec = new EventCartridge();
```
mpkorstanje: That would put the EventCartridge and EscapeHtmlReference in a scope that is greater then strictly needed. Neither is expensive to create nor created often so I don't see a good performance reason either.

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
mpkorstanje: Ok.

src/main/resources/templates/macros/json/embeddings.vm
mpkorstanje: Np. Just a formatter setting.

src/test/java/net/masterthought/cucumber/generators/integrations/FailuresOverviewPageIntegrationTest.java
mpkorstanje: Check.

src/test/java/net/masterthought/cucumber/json/FeatureTest.java
mpkorstanje: Check.

src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
mpkorstanje: Ok. Will rebase on next PR.

The package order come from the default settings. In .idea. I'll see if I can find that setting.

src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
mpkorstanje: If you drill down deep enough into `getDescription` you'll see that it uses `TextNode.getWholeText()` which uses the [html text attribute](http://www.w3schools.com/jsref/prop_node_textcontent.asp). This returns the textual representation of the nodes content. Because it is the textual representation any escaped characters are no longer escaped. As it should be, because this is what the user sees.

src/test/java/net/masterthought/cucumber/generators/integrations/helpers/EmbeddingAssertion.java
```diff
@@ -19,7 +19,8 @@ public void hasImageContent(Embedding embedding) {
     }
 
     public void hasTextContent(String content) {
-        assertThat(getBox().oneBySelector("pre", WebAssertion.class).html()).isEqualTo(getDecodedData(content));
+        assertThat(getBox().oneBySelector("pre", WebAssertion.class).text())
+                .isEqualTo(getDecodedData(content));
     }
```
mpkorstanje: Replaced .html with .text so the decoded data doesn't need to be escaped to match up.

src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
```diff
@@ -48,6 +49,33 @@ public void generateReport_CreatesReportFile() {
...
+                .isEqualTo("Error message java.lang.AssertionError: \n" +
+                        "Expected: is <80>\n" +
+                        "     got: <90>\n" +
+                        "\n" +
+                        "\tat org.junit.Assert.assertThat(Assert.java:780)\n" +
+                        "\tat org.junit.Assert.assertThat(Assert.java:738)\n" +
+                        "\tat net.masterthought.example.ATMScenario.checkBalance(ATMScenario.java:69)\n" +
+                        "\tat ✽.And the account balance should be 90(net/masterthought/example/ATMK.feature:12)");
+    }
+
```
mpkorstanje: Best I can make of it. Works for the purpose of verifying that html is escaped.

mpkorstanje:
Rebased and edits squashed.

src/test/java/net/masterthought/cucumber/generators/EscapeHtmlReferenceTest.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public class EscapeHtmlReferenceTest {
+
+    private static final String SOME_REFERENCE = "someReference";
+    private final ReferenceInsertionEventHandler insertionEventHandler = new EscapeHtmlReference();
+
+    @Test
+    public void referenceInsert_returnNormalText(){
+        String normalText = "a plain statement";
+        assertThat(insertionEventHandler.referenceInsert(SOME_REFERENCE, normalText)).isEqualTo(normalText);
```
damianszczepanik: I prefer https://en.wikipedia.org/wiki/Behavior-driven_development as the way how tests are implemented

src/test/java/net/masterthought/cucumber/generators/EscapeHtmlReferenceTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @Test
+    public void referenceInsert_shouldEscapeHtmlForAnyLabel(){
+        String html = "<b>a bold statement</b>";
+        assertThat(insertionEventHandler.referenceInsert(SOME_REFERENCE, html)).isEqualTo(escapeHtml(html));
+    }
+
+    @Test
+    public void referenceInsert_shouldReturnNullForNull(){
+        assertThat(insertionEventHandler.referenceInsert(SOME_REFERENCE, null)).isNull();
```
damianszczepanik: I'm happy you covered all cases

src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
```diff
@@ -48,6 +49,33 @@ public void generateReport_CreatesReportFile() {
...
+        // when
+        page.generatePage();
+
+        // then
+        DocumentAssertion document = documentFrom(page.getWebPage());
+        assertThat(document.getFeature().getDescription())
+                .isEqualTo("As an Account Holder I want to withdraw cash from an ATM,<br>so that I can get money when the bank is closed");
+        assertThat(document.getFeature().getElements()[0].getStepsSection().getSteps()[5].getEmbedding()[3].text())
+                .isEqualTo("Attachment 4 (HTML) <i>Hello</i> <b>World!</b>");
+        assertThat(document.getFeature().getElements()[0].getStepsSection().getSteps()[5].getMessage().text())
```
damianszczepanik: text() could be hidden inside of getMessage()

damianszczepanik:
Good job

damianszczepanik:
@mpkorstanje

There is regression - charts for tags and trends are not displayed

mpkorstanje:
Taking a look. Probely missed some more html injection.

mpkorstanje:
Fixed with #584.

PR #580, opened 2016-12-22
Fixes #579: Embedding links don't work correctly when same identical …

### LINKED ISSUE #579: Embedding links don't work correctly when same identical content is used
Hi @damianszczepanik,

We have observed that in a report with multiple scenarios, each having its own embeddings, when the string content of the embedding is identical, clicking on embedding links expands/collapses only the first embedding in the series.

For example, when you have a feature file:
```
Feature: Examples

Scenario: Scenario 1
   Given ...
   When ...
   Then ...

Scenario: Scenario 2
   Given ...
   When ...
   Then ...

Scenario: Scenario 3
   Given ...
   When ...
   Then ...
```

If all three scenarios generate same embedding, then when viewing the report for the feature file, and clicking on a link for expanding embedding number 2 or 3, it would expand/collapse embedding from scenario 1. 

This is caused by the fact that fileId is generated based on String's hashcode() method which returns same hashCode for identical strings. 

I have forked this repo, created a test and added a fix. I will reference this issue from the PR.

sponte:
…content is used

Issue #579

### DIFF
--- a/src/main/java/net/masterthought/cucumber/json/Embedding.java
+++ b/src/main/java/net/masterthought/cucumber/json/Embedding.java
@@ -1,18 +1,20 @@
 package net.masterthought.cucumber.json;
 
-import java.nio.charset.StandardCharsets;
-
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import net.masterthought.cucumber.json.deserializers.EmbeddingDeserializer;
 import org.apache.commons.lang.StringEscapeUtils;
 import org.codehaus.plexus.util.Base64;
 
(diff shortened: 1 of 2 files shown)


damianszczepanik:
I cannot reproduce the problem, can you attach JSON file?

sponte:
Hi @damianszczepanik, turns out it was fixed in 3.2.0 and I had a problem with 3.1.0 in our Jenkins instance. The fix is in templates/macros/report/embeddings.vm in includeExpandableEmbedding where you don't use getFileId() and instead use internal counter. I will close this PR.

Thanks for getting back to me.

PR #584, opened 2016-12-29
Render JS using the velocity engine

mpkorstanje:
It is preferable to render js through the template engine over rendering it at in java. This ensures that any transformations happen as late as possible


mpkorstanje:
Please think about renaming the other javascript files to .js.vm. I had not looked at the javascript files because I assumed to be static and not velocity templates.

src/main/resources/templates/js/trends-chart.js.vm
```diff
@@ -1,20 +1,23 @@
+#macro(stringArray $array) [#foreach($element in $array) "$element", #end] #end
+#macro(numberArray $array) [#foreach($element in $array) $element, #end] #end
+
```
mpkorstanje: Trailing comma's in lists are okay in javascript.

src/main/resources/templates/generators/overviewTrends.vm
```diff
@@ -5,7 +5,7 @@
   #includeTitle("Trends Overview")
 
   <script type="text/javascript">
-  #parse("/templates/js/trends-chart.js")
+  #parse("/templates/js/trends-chart.js.vm")
   </script>
```
mpkorstanje: Renamed the file for clarity - it should be explicit that this is a template.

src/main/resources/templates/js/trends-chart.js.vm
```diff
@@ -1,20 +1,23 @@
...
-    var pendingSteps = $pendingSteps;
-    var undefinedSteps = $undefinedSteps;
+    var passedSteps = #numberArray($passedSteps);
+    var failedSteps = #numberArray($failedSteps);
+    var skippedSteps = #numberArray($skippedSteps);
+    var pendingSteps = #numberArray($pendingSteps);
+    var undefinedSteps = #numberArray($undefinedSteps);
 
-    var durations = $durations;
+    var durations = #numberArray($durations);
```
mpkorstanje: number array and string array are not strictly needed, javascript will coerce  strings that represent numbers into numbers when needed but this just feels better, more precise.

damianszczepanik:
It does not solve the main regression introduced in pull #576

src/main/resources/templates/generators/overviewTrends.vm
```diff
@@ -5,7 +5,7 @@
   #includeTitle("Trends Overview")
 
   <script type="text/javascript">
-  #parse("/templates/js/trends-chart.js")
+  #parse("/templates/js/trends-chart.js.vm")
```
damianszczepanik: I don't like this change because of:
- now some files are js, others vm
- by this change IDEA does not color syntax by default

src/main/resources/templates/js/trends-chart.js.vm
damianszczepanik: I prefer to keep logic in Java because it allows me to test it easily and by having more logic in velocity I'm increasing potential regression like the one with tags because this macro is not tested

mpkorstanje:
Works for me:

works for me

Can you verify that you get this as the output:
overview-trends html

src/main/resources/templates/generators/overviewTrends.vm
mpkorstanje: 1. You should also rename the other js files where they are used as templates but that is outside the scope of this change. 

2. Hate to do this. But it works for me:

trends-charts-js vm

src/main/resources/templates/js/trends-chart.js.vm
mpkorstanje: That is architecturally not a sensible decision. The template engine is responsible for transforming the data into the output format and knows what context it will be injected into. Sidestepping this by using java adds complexity as java does not know the output format or any restrictions that apply such as the need to escape characters.

I will look into a test to see if the correct lines are output into overview-trends.html but frankly said, this part of the code is not setup for automated testing and will always need a manual test.

mpkorstanje:
Additional test added to TrendsOverviewPageIntegrationTest

mpkorstanje:
Found more manual array creation in the TagsOverviewPage.

mpkorstanje:
TagsOverviewPage is fixed.

works2

src/main/resources/templates/js/tags-chart.js.vm
```diff
@@ -1,32 +1,35 @@
+#macro(stringArray $array) [#foreach($element in $array) "$element", #end] #end
+#macro(numberArray $array) [#foreach($element in $array) $element, #end] #end
+
```
mpkorstanje: Duplicated this. Best way to dedupe it?

src/main/java/net/masterthought/cucumber/generators/TagsOverviewPage.java
```diff
@@ -38,14 +36,14 @@ public void prepareReport() {
...
     }
 
-    static String generateTagLabels(List<TagObject> tagsObjectList) {
+    static String[] generateTagLabels(List<TagObject> tagsObjectList) {
         int tagCount = tagsObjectList.size();
         String[] tagNames = new String[tagCount];
 
         for (int i = 0; i < tagCount; i++) {
-            tagNames[i] = StringUtils.wrap(tagsObjectList.get(i).getName(), "\"");
+            tagNames[i] = tagsObjectList.get(i).getName();
```
damianszczepanik: so now you have some logic here and also in macro.js.vm file what is even worse than now :(

src/main/resources/templates/generators/overviewTags.vm
```diff
@@ -5,7 +5,7 @@
   #includeTitle("Tags Overview")
 
   <script type="text/javascript">
-  #parse("/templates/js/tags-chart.js")
+  #parse("/templates/js/tags-chart.js.vm")
```
damianszczepanik: if you want to change it to vm then apply this change for all files - after your change some JavaScript charts are in js files, others in vm

src/main/resources/templates/macros/array.js.vm
```diff
@@ -0,0 +1,2 @@
+#macro(stringArray $array) [#foreach($element in $array) "$element", #end] #end
```
damianszczepanik: name of the file tells developer that there is some JavaScript code here - there is no

src/test/java/net/masterthought/cucumber/generators/TagsOverviewPageTest.java
```diff
@@ -66,13 +66,14 @@ public void prepareReport_setTagsToExcludeFromChart_ReturnsFilteredTags() {
...
         // then
         VelocityContext context = page.context;
-        assertThat(context.get("chart_categories")).isEqualTo("[\"@fast\"]");
-        assertThat(context.get("chart_data")).isEqualTo(asList(
-                "[100.00]",
-                "[0.00]",
-                "[0.00]",
-                "[0.00]",
-                "[0.00]"));
+        assertThat(context.get("chart_categories")).isEqualTo(new String[]{"@fast"});
```
damianszczepanik: there is probably method to extract element from that array and then compare the first one so `new String[]` is not needed any more

src/test/java/net/masterthought/cucumber/generators/TagsOverviewPageTest.java
```diff
@@ -66,13 +66,14 @@ public void prepareReport_setTagsToExcludeFromChart_ReturnsFilteredTags() {
...
+        assertThat(context.get("chart_data")).isEqualTo(new String[][]{
+                {"100.00"},
```
damianszczepanik: same

src/test/java/net/masterthought/cucumber/generators/TagsOverviewPageTest.java
```diff
@@ -82,10 +83,10 @@ public void generateTagLabels_ReturnsTags() {
         List<TagObject> allTags = this.tags;
 
         // when
-        String labels = TagsOverviewPage.generateTagLabels(allTags);
+        String[] labels = TagsOverviewPage.generateTagLabels(allTags);
 
         // then
-        assertThat(labels).isEqualTo("[\"@checkout\",\"@fast\",\"@featureTag\"]");
+        assertThat(labels).isEqualTo(new String[]{"@checkout","@fast","@featureTag"});
```
damianszczepanik: same

src/test/java/net/masterthought/cucumber/generators/TagsOverviewPageTest.java
```diff
@@ -95,14 +96,15 @@ public void generateTagValues_ReturnsTagValues() {
...
+        String[][] labels = TagsOverviewPage.generateTagValues(allTags);
 
         // then
-        assertThat(labels).containsExactly(
-                "[62.50, 100.00, 100.00]",
-                "[6.25, 0.00, 0.00]",
-                "[12.50, 0.00, 0.00]",
-                "[6.25, 0.00, 0.00]",
-                "[12.50, 0.00, 0.00]");
+        assertThat(labels).isEqualTo(new String[][]{
```
damianszczepanik: I believe there is some containsExactly for two dimentional arrays - take a look if it won't be simpler

src/test/java/net/masterthought/cucumber/generators/TrendsOverviewPageTest.java
```diff
@@ -66,61 +66,19 @@ public void prepareReport_AddsCustomProperties() {
         VelocityContext context = page.context;
         assertThat(context.getKeys()).hasSize(18);
 
-        assertThat(context.get("buildNumbers")).isEqualTo("[\"01_first\",\"other build\",\"05last\"]");
-        assertThat(context.get("failedFeatures")).isEqualTo("[1,2,5]");
-        assertThat(context.get("totalFeatures")).isEqualTo("[10,20,30]");
+        assertThat(context.get("buildNumbers")).isEqualTo(new String[]{"01_first","other build","05last"});
```
damianszczepanik: same

src/test/java/net/masterthought/cucumber/generators/integrations/TagsOverviewPageIntegrationTest.java
```diff
@@ -70,6 +70,30 @@ public void generatePage_generatesCharts() {
...
+        // check that data used by the charts is correctly inserted into the script section
+
+        assertThat(document.html()).contains("labels:  [ \"@checkout\",  \"@fast\",  \"@featureTag\", ]");
+        assertThat(document.html()).contains("data:  [ 62.50,  100.00,  100.00, ]");
+        assertThat(document.html()).contains("data:  [ 6.25,  0.00,  0.00, ]");
+        assertThat(document.html()).contains("data:  [ 12.50,  0.00,  0.00, ]");
+        assertThat(document.html()).contains("data:  [ 6.25,  0.00,  0.00, ]");
+        assertThat(document.html()).contains("data:  [ 12.50,  0.00,  0.00, ]");
+
+    }
```
damianszczepanik: nice

src/test/java/net/masterthought/cucumber/generators/integrations/TrendsOverviewPageIntegrationTest.java
```diff
@@ -75,4 +75,33 @@ public void generatePage_generatesCharts() {
...
+        assertThat(document.html()).contains("var totalFeatures =  [ 10,  20,  30, ] ;");
+        assertThat(document.html()).contains("var failedScenarios =  [ 10,  20,  20, ] ;");
+        assertThat(document.html()).contains("var totalScenarios =  [ 10,  2,  5, ] ;");
+        assertThat(document.html()).contains("var passedSteps =  [ 1,  3,  5, ] ;");
+        assertThat(document.html()).contains("var failedSteps =  [ 10,  30,  50, ] ;");
+        assertThat(document.html()).contains("var skippedSteps =  [ 100,  300,  500, ] ;");
+        assertThat(document.html()).contains("var pendingSteps =  [ 1000,  3000,  5000, ] ;");
+        assertThat(document.html()).contains("var undefinedSteps =  [ 10000,  30000,  50000, ] ;");
+        assertThat(document.html()).contains("var durations =  [ 3206126182398,  3206126182399,  3206126182310, ] ;");
+
```
damianszczepanik: very nice

src/main/resources/templates/macros/array.js.vm
mpkorstanje: This will render 

    ["something", "something", "something",] 

which is a javascript array. Unsure what I should call it.

src/test/java/net/masterthought/cucumber/generators/TagsOverviewPageTest.java
mpkorstanje: There is IntArrayAssert that allows these operations to work on arrays but is these are created through: `assertThat(int[] actual)`. Because `context.get()` always returns an object, the method binds to  `assertThat(T actual)` and we don't get that fancy matcher. .

mpkorstanje: Can do a cast and then use containsExactly but that doesn't improve readability.  Example:

```java
        assertThat((String[]) context.get("buildNumbers")).containsExactly("01_first", "other build", "05last");
        assertThat((int[]) context.get("failedFeatures")).containsExactly(1, 2, 5);
        assertThat((int[]) context.get("totalFeatures")).containsExactly(10, 20, 30);

        assertThat((int[]) context.get("failedScenarios")).containsExactly(10, 20, 20);
        assertThat((int[]) context.get("totalScenarios")).containsExactly(10, 2, 5);

        assertThat((int[]) context.get("passedSteps")).containsExactly(1, 3, 5);
        assertThat((int[]) context.get("failedSteps")).containsExactly(10, 30, 50);
        assertThat((int[]) context.get("skippedSteps")).containsExactly(100, 300, 500);
        assertThat((int[]) context.get("pendingSteps")).containsExactly(1000, 3000, 5000);
        assertThat((int[]) context.get("undefinedSteps")).containsExactly(10000, 30000, 50000);

        assertThat((long[]) context.get("durations")).containsExactly(3206126182398L, 3206126182399L, 3206126182310L);
```

src/test/java/net/masterthought/cucumber/generators/TagsOverviewPageTest.java
mpkorstanje: Makes things worse

```java
  // then
        assertThat(labels).containsExactly(
                new String[]{"62.50", "100.00", "100.00"},
                new String[]{"6.25", "0.00", "0.00"},
                new String[]{"12.50", "0.00", "0.00"},
                new String[]{"6.25", "0.00", "0.00"},
                new String[]{"12.50", "0.00", "0.00"}
        );
```

src/main/java/net/masterthought/cucumber/generators/TagsOverviewPage.java
mpkorstanje: Previously the function had two responsibilities. 1) transform the the list of objects into an array of strings and 2) translate that array of strings into a string that represents a javascript array of strings. The latter was also done in two different places in two different ways.  This has been split into two functions and the redundancy has been removed.

Both the transformation and the translation are now tested from end to end which wasn't done before as the tests did not look at the generated javascript. So in all I reckon we're up from where we were.

src/main/resources/templates/generators/overviewTags.vm
mpkorstanje: Ok.

PR #518, opened 2016-09-14
Added output messages grouping

v1nt06:
Hi Damian,

I made some changes in code after previous pull request. I added different output templates for outputs from different sources. But functional changes are the same:
- Added grouping for output messages inside steps
- Added word wrapping for messages and errors
- Error message title contains description (1st line of full error message)
- Error message is expanded by default

Before:
before

After:
after

Can you accept that pull request?
[...]

PR #594, opened 2017-01-25
Output messages grouping

v1nt06:
Hello Damian,

About 3 months ago I created same pull request, but I was busy to fix changes after your review ([previous pull request with my new replies](https://github.com/damianszczepanik/cucumber-reporting/pull/518#pullrequestreview-3009123)). So, now I complete this and rewrote some code. Could you take a look and accept pull request if everything is good?  Current functional changes in my fork:

- Added grouping for output messages inside steps
- Outputs from different sources grouped separately
- Added word wrapping for messages and errors
- Error message title contains description (1st line of full error message, e.g. exception name)
- Error messages and outputs is expanded by default

Sample.json with this functional changes:
sample json

Test scenario with different outputs:
expample


src/main/java/net/masterthought/cucumber/json/deserializers/StepDeserializer.java
```diff
@@ -0,0 +1,65 @@
...
+import net.masterthought.cucumber.json.Match;
+import net.masterthought.cucumber.json.Output;
+import net.masterthought.cucumber.json.Result;
+import net.masterthought.cucumber.json.Row;
+import net.masterthought.cucumber.json.Step;
+
+public class StepDeserializer extends JsonDeserializer<Step> {
+
+	@Override
+	public Step deserialize(JsonParser parser, DeserializationContext context) throws IOException {
```
damianszczepanik: instead of providing deserializer that makes code ugly, maybe try to provide additional method such as getOutputs() that wraps it? Then the model stays the  same as iti s in json file and onlyl view is using method for displaying content in different way

src/main/resources/templates/macros/json/output.vm
```diff
@@ -1,16 +1,21 @@
...
+      #set($outputId = $counter.next())
+      #set($index = $foreach.index + 1)
       <div class="output indention">
-        #set($outputId = $counter.next())
-        <div data-toggle="collapse" class="collapsable-control" data-target="#msg-$outputId">
-          #set($index = $foreach.index + 1)
-          <a>Output $index</a>
+        <div data-toggle="collapse" class="#if ($isPassed) collapsed #end collapsable-control" data-target="#output-$outputId">
+          Output $index
+          <i class="chevron fa fa-fw"></i>
```
damianszczepanik: what is it used for?

src/main/resources/templates/macros/json/output.vm
```diff
@@ -1,16 +1,21 @@
-#macro(includeOutput, $output)
+#macro(includeOutput, $outputs, $isPassed)
 
-#if ($output)
+#if($outputs)
   <div class="outputs inner-level">
-    #foreach($message in $output.getMessages())
+    #foreach($output in $outputs)
```
damianszczepanik: provided screenshtos don't really shows what is your intention. Can you describe it?

src/test/java/net/masterthought/cucumber/json/StepTest.java
```diff
@@ -65,10 +65,10 @@ public void getOutput_ReturnsOutput() {
         Step step = features.get(1).getElements()[0].getSteps()[7];
 
         // when
-        Output output = step.getOutput();
+        Output[] outputs = step.getOutputs();
 
         // then
-        assertThat(output.getMessages()).containsOnlyOnce(
+        assertThat(getMessages(outputs)).containsOnlyOnce(
```
damianszczepanik: finally all commits should be squals into one

src/main/java/net/masterthought/cucumber/json/deserializers/StepDeserializer.java
v1nt06: We can't wraps it because in this case we can't determine what was in json. For example, we can't determine what was in json: two diffirent outputs with one message in each or one output with two messages. So we need to detect this situations in deserialization process.

src/main/resources/templates/macros/json/output.vm
v1nt06: Sure. Here is demo screenshots.
Current output:
old

New output:
new

src/main/resources/templates/macros/json/output.vm
v1nt06: For collapsing functionality for outputs at the same style like steps. Please see screenshots above.

src/test/java/net/masterthought/cucumber/json/StepTest.java
v1nt06: Ok.

v1nt06:
All changes in one commit was added above. (4c035a9)

src/main/java/net/masterthought/cucumber/json/Step.java
```diff
@@ -1,22 +1,33 @@
...
-    @JsonProperty("doc_string")
-    private final DocString docString = null;
-    // End: attributes from JSON file report
+    private String name;
+    private final String keyword;
+    private final Result result;
+    private final Row[] rows;
+    private final Match match;
+    private final Embedding[] embeddings;
+    private final Output[] outputs;
```
damianszczepanik: have you tried to serialize using Object type? then the rest is used as now and the only implementation you need to provide is for single or array outputs

AFAIR this is how it was deserialized at the very beggining

src/main/java/net/masterthought/cucumber/json/deserializers/StepDeserializer.java
```diff
@@ -0,0 +1,65 @@
...
+        Result result = mapper.convertValue(rootNode.get("result"), Result.class);
+        Row[] rows = mapper.convertValue(rootNode.get("rows"), Row[].class) == null ?
+                new Row[0] : mapper.convertValue(rootNode.get("rows"), Row[].class);
+        Match match = mapper.convertValue(rootNode.get("match"), Match.class);
+        Embedding[] embeddings = mapper.convertValue(rootNode.get("embeddings"), Embedding[].class);
+        DocString docString = mapper.convertValue(rootNode.get("doc_string"), DocString.class);
+        Output[] outputs = parseOutputs(rootNode.get("output"));
+
+        return new Step(name, keyword, result, rows, match, embeddings, docString, outputs);
+    }
```
damianszczepanik: I want to minimalize code and execution that can be performed by 3td party lib

src/test/java/net/masterthought/cucumber/generators/integrations/PageTest.java
```diff
@@ -38,4 +42,13 @@ protected DocumentAssertion documentFrom(String pageName) {
             throw new ValidationException(e);
         }
     }
+
+    protected String[] getMessages(Output[] outputs) {
+        List<String> messages = new ArrayList<>();
+        for (Output output : outputs) {
+            messages.addAll(Arrays.asList(output.getMessages()));
```
damianszczepanik: I don't see unit test for this code

src/test/java/net/masterthought/cucumber/json/StepTest.java
damianszczepanik: generally I'm OK with this feature as long as this does not generate more code that is needed...

src/main/java/net/masterthought/cucumber/json/Step.java
v1nt06: I returned previous implementation and created deserializer only for output node.

src/main/java/net/masterthought/cucumber/json/deserializers/StepDeserializer.java
v1nt06: No longer relevant. This code was deleted after creating deserializer only for output node.

src/test/java/net/masterthought/cucumber/generators/integrations/PageTest.java
v1nt06: This method used only in test and designed for getting output messages for step tests and feature report page integration tests.

v1nt06:
Damian, I've pushed new changes. Could you take a look please?

src/main/java/net/masterthought/cucumber/json/Result.java
```diff
@@ -28,4 +28,8 @@ public String getFormatedDuration() {
     public String getErrorMessage() {
         return errorMessage;
     }
+
+    public final String getErrorMessageTitle() {
+        return errorMessage.split("[\\r\\n]+")[0];
```
damianszczepanik: Based on https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html
\s should be enough

src/main/java/net/masterthought/cucumber/json/Step.java
```diff
@@ -13,7 +15,9 @@
     private final Row[] rows = new Row[0];
     private final Match match = null;
     private final Embedding[] embeddings = new Embedding[0];
-    private final Output output = null;
+    @JsonDeserialize(using = OutputsDeserializer.class)
+    @JsonProperty("output")
+    private final Output[] outputs = new Output[0];
```
damianszczepanik: Have you tried to use single class and deserialilze it inside? so line 19 won't be required and deserializer method will be linked iwth Outputs class. Then also there will be single attribute here Objects, not the array

So it is more like the current solution

src/main/resources/templates/macros/json/output.vm
```diff
@@ -1,16 +1,21 @@
...
+      #set($outputId = $counter.next())
+      #set($index = $foreach.index + 1)
```
damianszczepanik: why not using counter as previously?

src/main/resources/templates/macros/json/output.vm
damianszczepanik: Looking at this I wonder why not having single output section?

src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
```diff
@@ -66,7 +66,7 @@ public void generateReport_DisplaysContentAsEscapedText() {
         assertThat(document.getFeature().getElements()[0].getStepsSection().getSteps()[5].getEmbedding()[3].text())
                 .isEqualTo("Attachment 4 (HTML) <i>Hello</i> <b>World!</b>");
         assertThat(document.getFeature().getElements()[0].getStepsSection().getSteps()[5].getMessage().text())
-                .isEqualTo("Error message java.lang.AssertionError: \n" +
+                .isEqualTo("java.lang.AssertionError: java.lang.AssertionError: \n" +
```
damianszczepanik: if the input data is not changed, why expected is?

src/main/java/net/masterthought/cucumber/json/Result.java
v1nt06: If we put into pattern "\s" we got only first word from message. Current pattern returns first line. For example, we have exception. In case with "\s" pattern we got only type of exception, but with current pattern we got type with message in one line. I think it is more useful.

src/main/resources/templates/macros/json/output.vm
v1nt06: Ok, I'll use only one counter.

v1nt06: > Looking at this I wonder why not having single output section?
Because we can't determine from wich output we got messages if we have few output sources. Do you mean that we need to group two different output sources into one like this? (two options):

1. 

2. 



> why not using counter as previously?
Because "index"-counter count output number in every step. "outputId"-counter need for collapsible sections.

src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
v1nt06: Input data was chnged in 14th line in steps.vm and in getErrorMessageTitle() inside Result class.

src/main/java/net/masterthought/cucumber/json/Step.java
v1nt06: No, I haven't tried. Because in this case we need to collect outputs (if we have more than 1 source) inside Output class in array and every element of this array should contain messages. But I can do this. What I should do?

src/main/resources/templates/macros/json/output.vm
damianszczepanik: Please try

damianszczepanik:
Rebase the code, squash commits and let me know how about this deserializer.

v1nt06:
Rebased and squashed. Deserializer remains as is. I wrote reason above:
https://github.com/damianszczepanik/cucumber-reporting/pull/594#discussion_r101259353

damianszczepanik:
I don't have more comments, thanks!

PR #597, opened 2017-02-01
Embed html in a sandboxed iframe

mpkorstanje:
It preferable to allow the browser to render embedded  html documents. To ensure this is done safely the html is placed in a sandboxed iframe. The html is embeded in the srcdoc as recommend by  the w3c[1]. The src attribute is provided as a backup for legacy browsers.

  Additionally because the cucumber-reporting-plugin must necessarily serve embedded html documents from the same domain as the report  the download link will now force a popup to open. While this can not prevent the embedded content from tricking the user into opening it directly, it at least mitigates some of the issues mentioned by the w3c[1].

   1. https://www.w3.org/TR/html5/embedded-content-0.html#the-iframe-element

### DIFF
--- a/src/main/resources/css/cucumber.css
+++ b/src/main/resources/css/cucumber.css
@@ -64,7 +64,7 @@ a:hover {
 /* same as <pre> from bootstrap library */
 .embedding-content {
     display: block;
-    padding: 9.5px;
+    padding: 10px;
     margin-left: 10px;
     margin-right: 10px;
     margin-bottom: 10px;
@@ -75,6 +75,23 @@ a:hover {
     border: 1px solid #ccc;
     border-radius: 4px;
     white-space: nowrap;
+}
+
+.html-content {
+    position: relative;
+    /* proportion value to aspect ratio 16:9 (9 / 16 = 0.5625 or 56.25%) */
+    padding: 0 0 56.25%;
+    height: 0;
+    overflow: hidden;
+}
+
+.html-content iframe {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    border:none;
 }
 
 .download-button {
--- a/src/main/resources/templates/macros/json/embeddings.vm
+++ b/src/main/resources/templates/macros/json/embeddings.vm
@@ -1,4 +1,4 @@
-#macro(includeEmbeddings, $embeddings)
+#macro(includeEmbeddings $embeddings)
 
   #if (!$embeddings.isEmpty())
   <div class="embeddings inner-level">
@@ -16,7 +16,7 @@
     #elseif ($embedding.getMimeType() == "text/plain")
       #includeTextEmbedding($embedding, "Plain text", $foreach.index)
     #elseif ($embedding.getMimeType() == "text/html")
-      #includeTextEmbedding($embedding, "HTML", $foreach.index)
+      #includeHtmlEmbedding($embedding, "HTML", $foreach.index)
     #elseif ($embedding.getMimeType() == "application/json")
       #includeTextEmbedding($embedding, "JSON", $foreach.index)
     #elseif ($embedding.getMimeType() == "image/url")
@@ -30,15 +30,35 @@
 
 #end
 
+#macro(includeHtmlEmbedding $embedding, $format, $index)
+  #set($index = $index + 1)
+<div class="embedding indention">
+  #set($embeddingId = $counter.next())
+  <div data-toggle="collapse" data-target="#embedding-$embeddingId" class="collapsable-control">
+    <a>Attachment $index ($format)</a>
+    <a href="embeddings/$embedding.getFileName()" download target="_blank">
+      <span class="download-button glyphicon glyphicon-download-alt"></span>
+    </a>
+  </div>
+  <div id="embedding-$embeddingId" class="collapse collapsable-details">
+    <div class="embedding-content">
+      <div class="html-content">
+        <iframe seamless="true" sandbox="allow-scripts" src="embeddings/$embedding.getFileName()" srcdoc="$embedding.getDecodedData()"></iframe>
+      </div>
+    </div>
+  </div>
+</div>
+#end
 
 
-#macro(includeImageEmbedding, $embedding, $image_type, $index)
+
+#macro(includeImageEmbedding $embedding, $image_type, $index)
   #set($index = $index + 1)
   <div class="embedding indention">
     #set($embeddingId = $counter.next())
     <div data-toggle="collapse" data-target="#embedding-$embeddingId" class="collapsable-control">
       <a>Attachment $index ($image_type)</a>
-      <a href="embeddings/$embedding.getFileName()" target="_blank">
+      <a href="embeddings/$embedding.getFileName()" download target="_blank">
         <span class="download-button glyphicon glyphicon-download-alt"></span>
       </a>
     </div>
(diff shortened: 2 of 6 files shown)


mpkorstanje:
Work in progress, open for discussion.

src/main/resources/templates/macros/json/embeddings.vm
```diff
@@ -1,4 +1,4 @@
-#macro(includeEmbeddings, $embeddings)
+#macro(includeEmbeddings $embeddings)
```
mpkorstanje: The comma isn't needed here.

src/main/resources/templates/macros/json/embeddings.vm
```diff
@@ -28,15 +28,33 @@
...
 
-#macro(includeImageEmbedding, $embedding, $image_type, $index)
+#macro(includeImageEmbedding $embedding, $image_type, $index)
   #set($index = $index + 1)
   <div class="embedding indention">
     #set($embeddingId = $counter.next())
     <div data-toggle="collapse" data-target="#embedding-$embeddingId" class="collapsable-control">
       <a>Attachment $index ($image_type)</a>
-      <a href="embeddings/$embedding.getFileName()" target="_blank">
+      <a href="embeddings/$embedding.getFileName()" download target="_blank">
```
mpkorstanje: Ensures the download dialog is triggered

src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
```diff
@@ -64,7 +64,7 @@ public void generateReport_DisplaysContentAsEscapedText() {
         assertThat(document.getFeature().getDescription())
                 .isEqualTo("As an Account Holder I want to withdraw cash from an ATM,<br>so that I can get money when the bank is closed");
         assertThat(document.getFeature().getElements()[0].getStepsSection().getSteps()[5].getEmbedding()[3].text())
-                .isEqualTo("Attachment 4 (HTML) <i>Hello</i> <b>World!</b>");
+                .isEqualTo("Attachment 4 (HTML)");
```
mpkorstanje: The element no longer contains any text. It is all contained in the srcDoc attribute of the iframe which we test in mbeddingAssertion.java

mpkorstanje:
@damianszczepanik  

can you fix sonar?

```ERROR] Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:3.2:sonar (default-cli) on project cucumber-reporting: You're not authorized to execute any SonarQube analysis. Please contact your SonarQube administrator. -> [Help 1]```

File: src/main/resources/css/cucumber.css [hunk 4]
```diff
@@ -77,6 +77,22 @@ a:hover {
...
+}
+
+.html-content iframe {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+}
+
```
mpkorstanje: This ensures the iframe can show its full height

damianszczepanik:
Don't worry about Sonar, The problem is with permission for pull request and I have some idea how to make it working again

damianszczepanik:
So the way how it works on Jenkins depends of the security level you have and the content (if your file refer to external resources or not) ?

mpkorstanje:
> So the way how it works on Jenkins depends of the security level you have and the content (if your file refer to external resources or not) ?

mpkorstanje:
Erh. Github ate my reply while hitting the wrong button. I'll try answering later to day.

mpkorstanje:
We mostly want to protect the Jenkins session cookie. The session cookie is available to any JavaScript served from the same domain e.g. buildserver.local.

Embedding the HTML in a sandbox ensures that it can not access the cookie. However if the user visits the actual file at buildserver.local/path/to/embeddings/untrusted.html the cookie is available because it is served from the same domain. It is currently trivial to make that happen. The download link does it.

The common solution is to serve the untrusted content from a subdomain like usercontent.buildserver. local. However this is not possible with Jenkins.

The next best solution would be to ensure the mimetype for the untrusted.html is not text/html so the browser won't interpret it but I am not familiar with the Jenkins plugin.

mpkorstanje:
@damianszczepanik 

I've been looking at the SafeArchiveServingAction in cucumber-reports-plugin. Am I correct in assuming this bypasses the [CSP](https://wiki.jenkins-ci.org/display/JENKINS/Configuring+Content+Security+Policy#ConfiguringContentSecurityPolicy-TheDefaultRuleSet)?

damianszczepanik:
That's correct, make sure you haven't dismissed CSP by enabling the flag before Jnkins is started

mpkorstanje:
Mmh. The embeddings folder should be served with CSP. But we can treat that as a separate issue.

damianszczepanik:
It does but mind that HTML file may need external resources and then user may be surprised that page is incomplete.

damianszczepanik:
I have fixed Sonar - can you rebase the code so I can verify if it was really fixed? Sonar should not run any more for pull requests.

mpkorstanje:
Done.

mpkorstanje:
Fixed the tests and the styling. Good enough as far as I am concerned.

mpkorstanje:
Regarding the CSP. We can not trust the embedded documents so if they're not allowed to do anything that is just fine. The embedded documents can load their resources when they're in the sandbox iframe so that should be alright.

damianszczepanik:
Agree, my question was: do we want to inform user that the content may be incomplete? I had several issues related to CSP so I would rather not have several false positives issues that the plugin does not properly display content. What do you think?

mpkorstanje:
1. The CSP should not apply to for example path/to/reports/report-feature_net-masterthought-example-ATMK-feature.html. We can trust that file because we created it and took care to neuter the untrusted embedded content.

No warning should be needed here.

2. The CSP should apply to path/to/reports/embeddings/embedding_123456790.html because that link should only be used to download the file. Viewing it poses a security risk.

A warning would be nice here, but we have no where to show it.

3. I'm going to setup a Jenkins for testing to see how it works. Just to assure myself I understand it all correctly.

damianszczepanik:
One more thing to do: modify sample.json file and add sample HTML page with external resouce (eg Wikipedia logo) so it will be easier to detect regression. Consider add test for this as well.

mpkorstanje:
No problem. Bit short on time at the moment.

mpkorstanje:
I have added sample html to the existing hello world with a promt telling the viewer the wikipedia logo should appear. FeatureReportPageIntegrationTest.generatePage_generatesEmbedding covers this by using hasSrcDocContent to check if the proper content has been inserted in the srcDoc attribute of the iframe.

image

Rebased onto latest head.

mpkorstanje:
I have also established that cucumber-reports jenkins plugin is not properly applying the CSP for embeddings. I will file a separate issue for that in the appropriate repository. This does not block this fix though. The problem existed before this PR and will exist after it.

Unless you have further comments I think this PR ready.

src/main/resources/css/cucumber.css
```diff
@@ -64,7 +64,7 @@ a:hover {
 /* same as <pre> from bootstrap library */
 .embedding-content {
     display: block;
-    padding: 9.5px;
+    padding: 10px;
```
mpkorstanje: No such thing as a half pixel.

src/test/java/net/masterthought/cucumber/generators/integrations/FeatureReportPageIntegrationTest.java
```diff
@@ -306,7 +306,7 @@ public void generatePage_generatesEmbedding() {
         asserEmbeddingFileExist(embeddings[0]);
         embeddingsElement[2].hasTextContent(embeddings[2].getData());
         asserEmbeddingFileExist(embeddings[2]);
-        embeddingsElement[3].hasTextContent(embeddings[3].getData());
+        embeddingsElement[3].hasSrcDocContent(embeddings[3].getData());
         asserEmbeddingFileExist(embeddings[3]);
```
mpkorstanje: Checks if the srcDoc attribute contains the embedded html.

damianszczepanik:
Good job!

mpkorstanje:
You are welcome!

PR #613, opened 2017-03-18
Santize trusted content

mpkorstanje:
Trusted content such as classifications may contain urls that should not be escaped.

Any reference prefixed with $\_sanitize_ will not be escaped but sanitized instead.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -262,6 +262,16 @@
             <artifactId>jsoup</artifactId>
             <version>1.10.1</version>
         </dependency>
+        <dependency>
+            <groupId>com.googlecode.owasp-java-html-sanitizer</groupId>
+            <artifactId>owasp-java-html-sanitizer</artifactId>
+            <version>20160924.1</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>21.0</version>
+        </dependency>
         <!-- for testing -->
         <dependency>
             <groupId>junit</groupId>
--- a/src/main/java/net/masterthought/cucumber/generators/EscapeHtmlReference.java
+++ b/src/main/java/net/masterthought/cucumber/generators/EscapeHtmlReference.java
@@ -2,16 +2,27 @@ package net.masterthought.cucumber.generators;
 
 import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.velocity.app.event.ReferenceInsertionEventHandler;
+import org.owasp.html.HtmlPolicyBuilder;
+import org.owasp.html.PolicyFactory;
 
 /**
  * Escapes all html and xml that was provided in a reference before inserting it into a template.
+ *
+ * References that start with $_sanitize_ will be sanitized to allow urls.
  */
 final class EscapeHtmlReference implements ReferenceInsertionEventHandler {
+
+    private static final PolicyFactory LINKS = new HtmlPolicyBuilder()
+            .allowStandardUrlProtocols().allowElements("a").allowAttributes("href")
+            .onElements("a").requireRelNofollowOnLinks().requireRelsOnLinks("noopener", "noreferrer")
+            .toFactory();
 
     @Override
     public Object referenceInsert(String reference, Object value) {
         if (value == null) {
             return null;
+        } else if(reference.startsWith("$_sanitize_")) {
+            return LINKS.sanitize(value.toString());
         } else {
             return StringEscapeUtils.escapeHtml(value.toString());
         }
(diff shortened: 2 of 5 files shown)


mpkorstanje:
I'll look at the failures later on.

pom.xml
```diff
@@ -262,6 +262,16 @@
             <artifactId>jsoup</artifactId>
             <version>1.10.1</version>
         </dependency>
+        <dependency>
+            <groupId>com.googlecode.owasp-java-html-sanitizer</groupId>
+            <artifactId>owasp-java-html-sanitizer</artifactId>
+            <version>20160924.1</version>
+        </dependency>
+        <dependency>
```
damianszczepanik: do we need ti? I don't see import for guava package

src/main/java/net/masterthought/cucumber/generators/EscapeHtmlReference.java
```diff
@@ -2,16 +2,27 @@
...
 
 /**
  * Escapes all html and xml that was provided in a reference before inserting it into a template.
+ *
+ * References that start with $_sanitize_ will be sanitized to allow urls.
  */
 final class EscapeHtmlReference implements ReferenceInsertionEventHandler {
 
+    private static final PolicyFactory LINKS = new HtmlPolicyBuilder()
+            .allowStandardUrlProtocols().allowElements("a").allowAttributes("href")
```
damianszczepanik: what if someone wants to display the image?

I think that it is defined by Jenkins administrator so we should allows for more than white list with single element that soon or later will be extended. What is your opinion?

src/test/java/net/masterthought/cucumber/generators/integrations/FeaturesOverviewPageIntegrationTest.java
```diff
@@ -58,8 +58,8 @@ public void generatePage_generatesLead() {
     public void generatePage_generatesClassifications() {
 
         // given
-        final String[] names = {"Platform", "Browser", "Branch"};
-        final String[] values = {"Win", "Opera", "master"};
+        final String[] names = {"Platform", "Browser", "Branch", "Repository"};
+        final String[] values = {"Win", "Opera", "master", "<a href=\"example.com\" rel=\"nofollow noopener noreferrer\">Example Repository</a>"};
```
damianszczepanik: nice

src/test/java/net/masterthought/cucumber/generators/integrations/helpers/TableRowAssertion.java
```diff
@@ -26,6 +26,16 @@ public LinkAssertion getReportLink() {
         return values;
     }
 
+    public String[] getCellsHtml() {
```
damianszczepanik: do we still need `getCellsValues` ?

pom.xml
mpkorstanje: Owasp depends on guava. It's listed in the parent pom of owasp as 
```
      <dependency>
        <groupId>com.google.guava</groupId>
        <artifactId>guava</artifactId>
        <!--
            We depend on 11 because it is the last which works with JDK 5
            which we want to link against.
        -->
        <version>[11.0,)</version>
        <scope>provided</scope>
      </dependency>
```

src/main/java/net/masterthought/cucumber/generators/EscapeHtmlReference.java
mpkorstanje: Can't allow everything. There are different levels of Jenkins administrator. 

I'm thinking it would be neater if we created an exhaustive list of data that can be used as classification and explicitly allow those.

Not that it should be very long list
 * Text
 * Links
 * Images

Thats about it I think.

src/test/java/net/masterthought/cucumber/generators/integrations/helpers/TableRowAssertion.java
mpkorstanje: FeatureReportPageIntegrationTest.generatePage_generatesArgument

damianszczepanik:
Looks good for the first version, we will observe what users expect. Thanks!

damianszczepanik:
@mpkorstanje after this commit some builds fail. Can you take and look?

mpkorstanje:
Check. #615 Picked the wrong version of Guava to use.

src/main/java/net/masterthought/cucumber/generators/EscapeHtmlReference.java
```diff
@@ -2,16 +2,27 @@
...
+            .onElements("a").requireRelNofollowOnLinks().requireRelsOnLinks("noopener", "noreferrer")
+            .toFactory();
+
     @Override
     public Object referenceInsert(String reference, Object value) {
         if (value == null) {
             return null;
+        } else if(reference.startsWith("$_sanitize_")) {
```
damianszczepanik: @mpkorstanje this two lines are not covered by tests. Would you like to add test for this?

PR #617, opened 2017-03-29
Show text embedding without superfluous borders

mpkorstanje:
Text and xml embedding were surounded by a border while other plain elements such as errors
and image attachment did not. This looked rather messy.

### DIFF
--- a/src/main/resources/css/cucumber.css
+++ b/src/main/resources/css/cucumber.css
@@ -61,9 +61,11 @@ a:hover {
     overflow: auto;
 }
 
-/* same as <pre> from bootstrap library */
+/*
+   same as <pre> from bootstrap library.
+   padding and overflow-x added.
+*/
 .embedding-content {
-    display: block;
     padding: 10px;
     margin-left: 10px;
     margin-right: 10px;
@@ -71,10 +73,12 @@ a:hover {
     font-size: 13px;
     overflow-x: auto;
     line-height: 1.42857143;
+    color: #333;
+    word-break: break-all;
+    word-wrap: break-word;
     background-color: #f5f5f5;
     border: 1px solid #ccc;
     border-radius: 4px;
-    white-space: nowrap;
 }
 
 .html-content {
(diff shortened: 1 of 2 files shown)


mpkorstanje:
Before:
plain-text-before

After:
after

mpkorstanje:
Urgh. Looks like I broke the line breaks in the xml.

mpkorstanje:
After 
after2

src/main/resources/css/cucumber.css
```diff
@@ -61,20 +61,24 @@ a:hover {
     overflow: auto;
 }
 
-/* same as <pre> from bootstrap library */
+/*
+   same as <pre> from bootstrap library.
+   padding and overflow-x added.
+*/
 .embedding-content {
-    display: block;
```
mpkorstanje: Replaced span with divs in embeddings.vm, so not needed.

src/main/resources/css/cucumber.css
```diff
@@ -61,20 +61,24 @@ a:hover {
...
     font-size: 13px;
     overflow-x: auto;
     line-height: 1.42857143;
+    color: #333;
+    word-break: break-all;
+    word-wrap: break-word;
     background-color: #f5f5f5;
     border: 1px solid #ccc;
     border-radius: 4px;
-    white-space: nowrap;
```
mpkorstanje: Pre handles this where needed.

src/main/resources/css/cucumber.css
```diff
@@ -61,20 +61,24 @@ a:hover {
...
     padding: 10px;
     margin-left: 10px;
     margin-right: 10px;
     margin-bottom: 10px;
     font-size: 13px;
     overflow-x: auto;
     line-height: 1.42857143;
+    color: #333;
+    word-break: break-all;
+    word-wrap: break-word;
```
mpkorstanje: These are in pre from bootstrap. Keeps sections that don't use pre consistent with sections that do.

damianszczepanik:
Thank you :)

PR #620, opened 2017-04-02
Add additional unit test

mpkorstanje:
Add unit tests for the sanitation case of the html escaping reference
handler.


damianszczepanik:
High test coverage is back - nice!

PR #622, opened 2017-04-05
ignoring steps with null results to support scenario outlines

bkitchener1:
This should be a fix for https://github.com/damianszczepanik/cucumber-reporting/issues/621

### DIFF
--- a/src/main/java/net/masterthought/cucumber/json/Element.java
+++ b/src/main/java/net/masterthought/cucumber/json/Element.java
@@ -126,6 +126,9 @@ public class Element implements Durationable {
         StatusCounter statusCounter = new StatusCounter();
         for (Step step : steps) {
             Result result = step.getResult();
+            if(result==null){
+                break;
+            }
             statusCounter.incrementFor(result.getStatus());
             duration += result.getDuration();
         }
--- a/src/main/java/net/masterthought/cucumber/json/Feature.java
+++ b/src/main/java/net/masterthought/cucumber/json/Feature.java
@@ -222,6 +222,9 @@ public class Feature implements Reportable, Durationable, Comparable<Feature> {
             }
 
             for (Step step : element.getSteps()) {
+                if(step.getResult() == null){
+                    break;
+                }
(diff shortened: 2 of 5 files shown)


src/main/java/net/masterthought/cucumber/json/Element.java
```diff
@@ -126,6 +126,9 @@ private Status calculateStepsStatus() {
         StatusCounter statusCounter = new StatusCounter();
         for (Step step : steps) {
             Result result = step.getResult();
+            if(result==null){
```
damianszczepanik: formatting

src/main/java/net/masterthought/cucumber/json/Element.java
```diff
@@ -126,6 +126,9 @@ private Status calculateStepsStatus() {
...
+                break;
```
damianszczepanik: for sure not ``break``

src/main/java/net/masterthought/cucumber/json/Feature.java
```diff
@@ -222,6 +222,9 @@ private void calculateSteps() {
             }
 
             for (Step step : element.getSteps()) {
+                if(step.getResult() == null){
+                    break;
+                }
```
damianszczepanik: same

src/test/java/net/masterthought/cucumber/ReportGenerator.java
```diff
@@ -18,6 +18,7 @@
     public final static String JSON_DIRECTORY = "json/";
 
     protected static final String SAMPLE_JSON = "sample.json";
+    protected static final String SCENARIO_OUTLINE = "scenario-outline.json";
```
damianszczepanik: I don't want to new test file, problematic cases should be merged with sample.json

src/main/java/net/masterthought/cucumber/json/Element.java
damianszczepanik: and then the StatusCounter.size() does not return proper value

damianszczepanik:
This was already fixed #627

PR #624, opened 2017-04-17
Santize span and img tags

tohaa:
Hello Damian,

I've made some changes in EscapeHtmlReference.java , it allows to hightlight text and add image to a report. Thus we escapes "span" and "img" html tags now. 
Could you take a look and accept pull request if everything is good?

### DIFF
--- a/src/main/java/net/masterthought/cucumber/generators/EscapeHtmlReference.java
+++ b/src/main/java/net/masterthought/cucumber/generators/EscapeHtmlReference.java
@@ -12,17 +12,17 @@ import org.owasp.html.PolicyFactory;
  */
 final class EscapeHtmlReference implements ReferenceInsertionEventHandler {
 
-    private static final PolicyFactory LINKS = new HtmlPolicyBuilder()
-            .allowStandardUrlProtocols().allowElements("a").allowAttributes("href")
-            .onElements("a").requireRelNofollowOnLinks().requireRelsOnLinks("noopener", "noreferrer")
+    private static final PolicyFactory TAGS = new HtmlPolicyBuilder()
+            .allowStandardUrlProtocols().allowElements("a", "span", "img").allowAttributes("href", "onclick", "src", "id", "style")
+            .onElements("a", "span", "img").requireRelNofollowOnLinks().requireRelsOnLinks("noopener", "noreferrer")
             .toFactory();
 
     @Override
     public Object referenceInsert(String reference, Object value) {
         if (value == null) {
             return null;
-        } else if(reference.startsWith("$_sanitize_")) {
-            return LINKS.sanitize(value.toString());
+        } else if(reference.contains("$_sanitize_")) {
+            return TAGS.sanitize(value.toString());
         } else {
             return StringEscapeUtils.escapeHtml(value.toString());
         }
--- a/src/main/java/net/masterthought/cucumber/generators/EscapeHtmlReference.java
+++ b/src/main/java/net/masterthought/cucumber/generators/EscapeHtmlReference.java
@@ -21,7 +21,7 @@ final class EscapeHtmlReference implements ReferenceInsertionEventHandler {
     public Object referenceInsert(String reference, Object value) {
         if (value == null) {
             return null;
-        } else if(reference.contains("$_sanitize_")) {
+        } else if(reference.startsWith("$_sanitize_")) {
             return TAGS.sanitize(value.toString());
         } else {
             return StringEscapeUtils.escapeHtml(value.toString());
--- a/src/main/resources/templates/macros/json/output.vm
+++ b/src/main/resources/templates/macros/json/output.vm
@@ -14,7 +14,7 @@
           #**
            * DO NOT format the line below. Whitespace nodes are significant in a pre-block.
            *#
-          <pre>#foreach($message in $output.getMessages())<p>$message</p>#end</pre>
+          <pre>#foreach($_sanitize_message in $output.getMessages())<p>$_sanitize_message</p>#end</pre>
         </div>
       </div>
     #end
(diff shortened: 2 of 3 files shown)


mpkorstanje:
If you want to use html in your report you should use:

```
String html = "some html";
scenario.embed(html.getBytes(), "text/html");
```

`scenario.write ` is for text only. Using html doesn't make sense as your tests do not -and should not- know what the reporter is capable on the other end of the chain.

src/main/java/net/masterthought/cucumber/generators/EscapeHtmlReference.java
```diff
@@ -12,17 +12,17 @@
  */
 final class EscapeHtmlReference implements ReferenceInsertionEventHandler {
 
-    private static final PolicyFactory LINKS = new HtmlPolicyBuilder()
-            .allowStandardUrlProtocols().allowElements("a").allowAttributes("href")
-            .onElements("a").requireRelNofollowOnLinks().requireRelsOnLinks("noopener", "noreferrer")
+    private static final PolicyFactory TAGS = new HtmlPolicyBuilder()
+            .allowStandardUrlProtocols().allowElements("a", "span", "img").allowAttributes("href", "onclick", "src", "id", "style")
+            .onElements("a", "span", "img").requireRelNofollowOnLinks().requireRelsOnLinks("noopener", "noreferrer")
             .toFactory();
```
mpkorstanje: `.allowAttributes("href", "onclick", "src", "id", "style").onElements("a", "span", "img")` 

This doesn't look right at all. Allowed attributes should be set for each individual element not for the whole group.

tohaa: Hello mpkorstanje,

embed - it's not a very good solution for highlighting text and highlighting warning messages. look at the screenshots please, there is highlighted text for test reproduction and a highlighted warning message

mpkorstanje:
Yes. I have seen the screenshots. You can and should do these things through `            scenario.embed(html.getBytes(), "text/html");`

screenshot from 2017-04-17 16-10-49

The problem of using html in scenario.write is that it is intended to be plain text an may end in places that do not support html such as the commandline.

For example when using the pretty printer what should look like this:

```
Feature: Example  
  Some things here to describe this feature 

Red and yellow and pink and green 
Purple and orange and orange 
I can sing a rainbow
Sing a rainbow
Sing a rainbow too

Listen with your eyes

  Scenario: Running an example with output # /home/mpkorstanje/Projects/cucumber/src/test/resources/features/integration.feature:5
    Given some situation                   # MyStepdefs.java:35
    When some action                       # MyStepdefs.java:35
    Then some result should happen         # MyStepdefs.java:35

1 Scenarios (1 passed)
3 Steps (3 passed)
0m0.122s

Process finished with exit code 0
```
Ends up looking like this.

```
Feature: Example  
  Some things here to describe this feature 

<html><body><span style="color:red">Red</span> and <span style="color:yellow ">yellow </span> and <span style="color:pink ">pink </span> and <span style="color:green ">green </span><br><span style="color:purple">Purple</span> and <span style="color:orange ">orange </span> and <span style="color:orange ">orange </span><br>I can sing a rainbow<br>Sing a rainbow<br>Sing a rainbow too<br><br>Listen with your eyes</body></html>

  Scenario: Running an example with output # /home/mpkorstanje/Projects/cucumber/src/test/resources/features/integration.feature:5


    Given some situation                   # MyStepdefs.java:35
    When some action                       # MyStepdefs.java:35
    Then some result should happen         # MyStepdefs.java:35

1 Scenarios (1 passed)
3 Steps (3 passed)
0m0.122s

Process finished with exit code 0
```

mpkorstanje:
This is a popular tool yes. And your use case is not the only one. It is perfectly reasonable for people to use html tags in `scenario.write` and expect to see those tags as plain text. Therefore it is important to be extremely conservative when interpreting the users output. 

For example rendering `scenario.write("...but the element contained <span>Lorum Ipsum</span>")` as `...but the element contained Lorum Ipsum` is going to lead to some confusion.

> On your screenshot, HTML text can't be embedded between logs and doesn't look good and readable as a result. 

What exactly do you mean by between logs? Do you mean that consecutive calls to `scenario.embed` are not displayed as a single block while consecutive calls to `scenario.write` are? I am certain you can find a way to work around this.

mpkorstanje:
On a historical note: That it was possible to inject html into the report should never have been considered a feature. It may have been useful but should have been recognized as an abuse of the reporting tool. So I do not think that bad decisions made in the past are a justification to add undue complexity to our tools now.

mpkorstanje:
@rommy1204, @tohaa 

I would recommend that you write an html logger. This is fairly trivial and you can easily extend it to log just about anything text, images, videos, ect.

```
public class HtmlLogger implements En {

    private List<String> logs = new LinkedList<>();

    public HtmlLogger() {
        After(scenario -> {
            scenario.embed(("<pre>" + join(logs, "\n") + "</pre>").getBytes(), "text/html");
            this.logs.clear();
        });
    }

    public void info(String msg){
        this.logs.add(line(INFO, msg));
    }

    public void severe(String msg){
        this.logs.add(line(SEVERE, msg));
    }

    private static String line(Level level, String msg) {
        String color = level.intValue() > INFO.intValue() ? "red" : "black";
        return "<span style=\"color:" + color + "\">" + level +  " " + escapeHtml4(msg)+ "</span>";
    }
}
```
Which you can then use as
```
public class MyStepdefs implements En {

    public MyStepdefs(HtmlLogger log) {

        Given("^some (.*)$", (String bla) -> {
            log.info(bla);
        });

        Given("^a severe warning$", () -> {
            log.severe("All out of fleeb juice!");
        });
    }
}
```

damianszczepanik:
I agree with @mpkorstanje who did great effort to make this report secure so it does not break Jenkins policy (I hope so!) so I'm not happy to make it less secure as long as there is no strong reason for this. Based on above conversation I don't see such reason.

PR #626, opened 2017-04-27
Update embeddings.vm

vszalontai:
I had to patch this to be able to display from JSON generated pngs using the Jenkins cucumber plugin... I may be tripping, but this seems to be the standard: https://en.wikipedia.org/wiki/Data_URI_scheme
It's up to you to accept this change.... how ever, embedding a base64 image using src=<path> doesn't work in any browser....

### DIFF
--- a/src/main/resources/templates/macros/json/embeddings.vm
+++ b/src/main/resources/templates/macros/json/embeddings.vm
@@ -64,7 +64,7 @@
     </div>
     <div id="embedding-$embeddingId" class="collapse collapsable-details">
       <div class="embedding-content">
-        <img src="embeddings/$embedding.getFileName()">
+        <img src="$embedding.getDecodedData()">
       </div>
     </div>
   </div>


damianszczepanik:
This change makes sense if user can download the file.

mpkorstanje:
It is possible to use data uri's in links as well. This allows the user to download the file and removes the need to store it separately. This will increase the size of the html file significantly though. 

Eventually if enough other changes are made this will allow cucumber reports to be a single page application that could in theory be emailed.

I am ambivalent about this change.

vszalontai:
Read this please: http://stackoverflow.com/questions/35828806/use-base64-string-from-url-in-src-tag-of-image
Means: embedding a picture using src="http://... or src="file:// using a velocity template will not work at all. You have to use JS or CSS to load the real decoded data OR just extract it into the src tag (that's what I did in this patch).
About sizes and performance:  in case of small images you save time, since everything can be transferred in one HTTP request. In case of big images, you loose performance, since the images can't be cached (this may depend on the browser and the embedding technique... CSS can be cashed in most browsers 4 example) or managed with CDN. It's a kind of complicated topic, however the current code (src="http://... or src="file://) will not work in any browser at all :)

mpkorstanje:
The current code works. And I am fairly certain that your patch does not. :)

You may want to check the embeddings folder that the report generates. The images are stored there.

vszalontai:
The current code "works", yes, the images are on the right place (inside the embeddings folder) BUT you will never get these pictures displayed in a browser...

mpkorstanje:
That too works fine. If you want to report a bug you should provide steps to reproduce the problem.

PR #644, opened 2017-07-19
Handle missing 'result' properties

zbstof:
We've implemented graceful shutdown with writing pretty json file even in case of interrupted test process.
Unfortunately, result property is missing from steps that haven't executed yet.

This change will allow us to generate html even for such json.

And, judging from the comment in https://github.com/damianszczepanik/cucumber-reporting/blob/master/src/main/java/net/masterthought/cucumber/json/Result.java#L13
we can just use Status.UNDEFINED without breaking the contract for Result.

Current behaviour is to throw
```
java.lang.NullPointerException
	at net.masterthought.cucumber.json.Element.calculateStepsStatus(Element.java:115)
	at net.masterthought.cucumber.json.Element.setMetaData(Element.java:91)
	at net.masterthought.cucumber.json.Feature.setMetaData(Feature.java:161)
	at net.masterthought.cucumber.ReportParser.setMetadata(ReportParser.java:97)
	at net.masterthought.cucumber.ReportParser.parseJsonFiles(ReportParser.java:60)
...

```
What do you think?

### DIFF
--- a/src/main/java/net/masterthought/cucumber/json/Element.java
+++ b/src/main/java/net/masterthought/cucumber/json/Element.java
@@ -1,10 +1,11 @@
 package net.masterthought.cucumber.json;
 
+import org.apache.commons.lang.StringUtils;
+
 import net.masterthought.cucumber.json.support.Durationable;
 import net.masterthought.cucumber.json.support.Status;
 import net.masterthought.cucumber.json.support.StatusCounter;
 import net.masterthought.cucumber.util.Util;
-import org.apache.commons.lang.StringUtils;
 
 public class Element implements Durationable {
 
--- a/src/main/java/net/masterthought/cucumber/json/Element.java
+++ b/src/main/java/net/masterthought/cucumber/json/Element.java
@@ -1,11 +1,10 @@
 package net.masterthought.cucumber.json;
 
-import org.apache.commons.lang.StringUtils;
-
 import net.masterthought.cucumber.json.support.Durationable;
 import net.masterthought.cucumber.json.support.Status;
 import net.masterthought.cucumber.json.support.StatusCounter;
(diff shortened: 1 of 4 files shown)


zbstof:
Related to https://github.com/damianszczepanik/cucumber-reporting/pull/622 :
Should this be done for Feature.java also?

src/main/java/net/masterthought/cucumber/json/Element.java
```diff
@@ -126,6 +126,9 @@ private Status calculateStepsStatus() {
         StatusCounter statusCounter = new StatusCounter();
         for (Step step : steps) {
             Result result = step.getResult();
+            if (result == null) {
+                result = new Result();
+            }
```
damianszczepanik: According to https://github.com/damianszczepanik/cucumber-reporting/commit/7d9b8c49e209d1a2d05937a5f4a4cd55cce63e0c#diff-1eeb2da981684473c263089457fbbca9R16 when this may happen?

src/test/java/net/masterthought/cucumber/ReportGenerator.java
```diff
@@ -20,6 +20,7 @@
 
     protected static final String SAMPLE_JSON = "sample.json";
     public static final String SIMPLE_JSON = "simple.json";
+    public static final String MISSING_RESULT_JSON = "missing-result.json";
```
damianszczepanik: I prefer to add sample.json file instead of adding new one

src/main/java/net/masterthought/cucumber/json/Element.java
zbstof: You are correct, it seems that was fixed in 3.9.0
My bad

PR #657, opened 2017-08-25
use undecoded data when embedded text is processed

timnederhoff:
Embeddings with mime_type text/plain, text/xml, and application/json are decoded as if it was base64. I disabled this so the text is readable in the html report.

### DIFF
--- a/src/main/resources/templates/macros/json/embeddings.vm
+++ b/src/main/resources/templates/macros/json/embeddings.vm
@@ -99,7 +99,7 @@
       </a>
     </div>
     <div id="embedding-$embeddingId" class="collapse collapsable-details">
-      <pre class="embedding-content">$embedding.getDecodedData()</pre>
+      <pre class="embedding-content">$embedding.getData()</pre>
     </div>
   </div>
 #end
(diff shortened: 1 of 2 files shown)


damianszczepanik:
Then the text is not displayed properly. Take a look at the result from sample json attached to the project and compare results

timnederhoff:
So text/plain, text/xml and application/json are supposed to be base64 encoded in the cucumber json? Is that the desired functionality? Or should there be a difference in processing these different input types?

damianszczepanik:
This is not the question to this project so I'm closing this request as there is no sense for now

mpkorstanje:
@timnederhoff 
> So text/plain, text/xml and application/json are supposed to be base64 encoded in the cucumber json?
Yes, all mimetypes are base64 encoded. See  [JSONFormatter](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/cucumber/runtime/formatter/JSONFormatter.java#L303) If they are not please file a bug report against your cucumber implementation.

timnederhoff:
Ah right thanks. I see that the Java JSONFormatter does encode, but the Javascript [attachment manager](https://github.com/cucumber/cucumber-js/blob/fbff6b0fae54d2e341ee247addc60a9f05753f1d/src/runtime/attachment_manager/index.js#L58) doesn't (the one I checked with).

PR #658, opened 2017-08-25
Support PDF embedding with `.pdf` extension

davigp:
Currently the PDF embedding is downloaded as "unknown":
download_unknown
With this commit the PDF embedding will be downloaded as "PDF document":
download_pdf


damianszczepanik:
Thanks!

PR #665, opened 2017-09-13
Failure Overview On Overview Features Page

seanBucholtz:
**Case:** many features (100+) each with one-to-many scenarios on the Overview Feature page with failures throughout the given set.

**Problem:** analyzing the cause of each failure across all features/scenarios becomes very time consuming when having to open the associated feature report to make the determination as to why each test case failed.

**Solution:** implement a quick _failure overview_ dropup to each feature row on the Overview Features page which lists the: scenario name, the step name/text (that links to the failing step in the associated feature report), and the error message that was encountered causing the step to fail. This enables the viewer to see exactly why each scenario failed without having to navigate away from the Overview Feature page, and if the individual would like further context, they can click the desired step link within the drop up which forwards them directly to the exact point of failure.

**Notes:** As per your contributing guidelines/requirements, I've expanded the unit and integration test to cover this new feature.

The quick _failure overview_ button can be seen in the Scenarios-Failed column.
featureoverviewpage

Upon clicking the quick _failure overview_ button, a dropup appears containing each failure.
featureoverviewpagedropup


damianszczepanik:
First, squash into single commit, then split into smaller parts so eg refactoring is not merged with other tests. Finally explain changes in model. Then I will be happy to discuss all of them

PR #672, opened 2017-10-20
Added support for the image/svg+xml image type

Hylke1982:
SVG images did not get the right extension and were not made available using the image tag.


damianszczepanik:
Squash commits into one then I will be happy to merge it

PR #673, opened 2017-10-20
Added support for the image/svg+xml image type

Hylke1982:
Added additional test for SVG images


damianszczepanik:
Sorry, one one kind request. Can you update https://github.com/damianszczepanik/cucumber-reporting/blob/master/src/test/resources/json/sample.json so there is at least one (even simple) svg attachment? It helps testing UI after changes and trigger regressions.

src/test/resources/json/sample.json
```diff
@@ -447,6 +447,10 @@
...
+                            },
+                            {
```
damianszczepanik: Great but now commits needs to be squashed again :)

Hylke1982:
I squashed the commit, but I see you have ran a maven prepare command on this branch. Should I squash this commit also?

mpkorstanje:
@damianszczepanik if I am not mistaken the merge button has a drop down now that lets you chose the merge mode. It includes squash and re-base. This can avoid some back and forth.

damianszczepanik:
Right, there is squash option on github - need to check what is the commit message then as I like when the commit says everything about change it delivers.

Second commit is now wrong, I believe you should rebasse the code and then it should gone

Hylke1982:
@damianszczepanik Rebased it and it should be releasable now.

damianszczepanik:
Thanks

PR #692, opened 2017-11-21
Allow undefined and pending steps to not fail

a1dutch:
Hi, This is a cut down version of #636, allows you to pass `strict` into the configuration object and makes the status `PASSED` as long as there are no `FAILURES`.  It would also mark `PENDING` as passed if strict was false too, but happy to make a change and fixup the imports if your willing to merge.

@a1dutch

### DIFF
--- a/src/main/java/net/masterthought/cucumber/Configuration.java
+++ b/src/main/java/net/masterthought/cucumber/Configuration.java
@@ -8,6 +8,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
+
 import net.masterthought.cucumber.sorting.SortingMethod;
 
 public class Configuration {
--- a/src/main/java/net/masterthought/cucumber/Configuration.java
+++ b/src/main/java/net/masterthought/cucumber/Configuration.java
@@ -8,7 +8,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
-
 import net.masterthought.cucumber.sorting.SortingMethod;
 
 public class Configuration {
@@ -17,6 +16,7 @@ public class Configuration {
 
     private boolean parallelTesting;
     private boolean runWithJenkins;
+    private boolean strict = true;
 
     private File reportDirectory;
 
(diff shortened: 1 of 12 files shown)


damianszczepanik:
Haven't seen details of the implementation but tell me the rules how the statuses should be changed and where the change will be visible, where not. I need to understand your point of view in advance.

src/main/java/net/masterthought/cucumber/Configuration.java
```diff
@@ -81,6 +83,20 @@ public int getTrendsLimit() {
         return trendsLimit;
     }
 
+    /**
+     * Sets whether to treat pending and undefined steps as failed.
```
damianszczepanik: I would like to have separate field for all three https://github.com/damianszczepanik/cucumber-reporting/blob/master/src/main/java/net/masterthought/cucumber/json/support/Status.java#L17-L19

src/main/java/net/masterthought/cucumber/json/Element.java
```diff
@@ -96,39 +96,39 @@ public String getFormattedDuration() {
         return Util.formatDuration(duration);
     }
 
-    public void setMetaData(Feature feature) {
+    public void setMetaData(Feature feature, boolean strict) {
```
damianszczepanik: ... and then pass here the configuration, instead of three different flags

src/main/java/net/masterthought/cucumber/json/support/StatusCounter.java
```diff
@@ -66,7 +56,10 @@ public int size() {
      *
      * @return final status for this counter
      */
-    public Status getFinalStatus() {
-        return finalStatus;
+    public Status getFinalStatus(boolean strict) {
+        if (strict) {
+            return counter.get(Status.PASSED) == size() ? Status.PASSED : Status.FAILED;
```
damianszczepanik: would you be able to see pending statuses on the overview pages if the `strict` mode is on?

src/test/java/net/masterthought/cucumber/json/support/StepObjectTest.java
```diff
@@ -34,7 +34,7 @@ public void StepObject_OnNullLocation_ThrowsException() {
 
         // then
         thrown.expect(ValidationException.class);
-        new StepObject(null);
+        new StepObject(null, true);
```
damianszczepanik: tests for no strict mode are missing

PR #694, opened 2017-11-24
Support screenshots in v2 and v3 reports

achingbrain:
Fixes jenkinsci/cucumber-reports-plugin#193 by conditionally trying to load the mime type of an embedding from the v3 location, if not found it will fall back to v2.  Also adds tests for the same.

Happy to make any amends necessary.

### DIFF
--- a/src/main/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializer.java
+++ b/src/main/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializer.java
@@ -24,12 +24,22 @@ public class EmbeddingDeserializer extends CucumberJsonDeserializer<Embedding> {
     @Override
     public Embedding deserialize(JsonNode rootNode, Configuration configuration) {
         String data = rootNode.get("data").asText();
-        String mimeType = rootNode.get("mime_type").asText();
+        String mimeType = findMimeType(rootNode);
 
         Embedding embedding = new Embedding(mimeType, data);
         storeEmbedding(embedding, configuration.getEmbeddingDirectory());
 
         return embedding;
+    }
+
+    private String findMimeType(JsonNode rootNode) {
+        JsonNode media = rootNode.get("media");
+
+        if (media != null) {
+            return media.get("type").asText();
+        }
+
+        return rootNode.get("mime_type").asText();
     }
 
     private void storeEmbedding(Embedding embedding, File embeddingDirectory) {
--- a/src/test/java/net/masterthought/cucumber/generators/FeatureReportPageTest.java
+++ b/src/test/java/net/masterthought/cucumber/generators/FeatureReportPageTest.java
@@ -2,6 +2,9 @@ package net.masterthought.cucumber.generators;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import net.masterthought.cucumber.json.Element;
+import net.masterthought.cucumber.json.Embedding;
+import net.masterthought.cucumber.json.Step;
 import org.apache.velocity.VelocityContext;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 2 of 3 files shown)


src/test/java/net/masterthought/cucumber/generators/FeatureReportPageTest.java
```diff
@@ -49,4 +52,25 @@ public void prepareReport_AddsCustomProperties() {
         assertThat(context.get("parallel")).isEqualTo(configuration.isParallelTesting());
         assertThat(context.get("feature")).isEqualTo(feature);
     }
+
+    @Test
```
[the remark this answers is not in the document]
damianszczepanik: Yes, I prefer to have //given //when //then ans the assertion instead of single test method

src/test/java/net/masterthought/cucumber/generators/FeatureReportPageTest.java
```diff
@@ -49,4 +52,25 @@ public void prepareReport_AddsCustomProperties() {
...
+    public void getMimeType_OnEmbeddingFromV2CucumberReportFile_SupportsScreenshots() {
+        testScreenshotsAreSupported(CUCUMBER_V2_WITH_SCREENSHOT_JSON);
+    }
+
+    @Test
+    public void getMimeType_OnEmbeddingFromV3CucumberReportFile_SupportsScreenshots() {
+        testScreenshotsAreSupported(CUCUMBER_V3_WITH_SCREENSHOT_JSON);
+    }
+
+    private void testScreenshotsAreSupported(String reportFile) {
```
damianszczepanik: can you use above test method naming convention?

src/test/resources/json/cucumber-v2-with-screenshot.json
```diff
@@ -0,0 +1,129 @@
+[
```
damianszczepanik: instead of adding new test json file add change you want to test into sample.json file

src/test/resources/json/cucumber-v3-with-screenshot.json
```diff
@@ -0,0 +1,134 @@
+[
```
damianszczepanik: if you have more than one case, merge them all into sample.json

src/main/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializer.java
```diff
@@ -24,7 +24,16 @@
     @Override
     public Embedding deserialize(JsonNode rootNode, Configuration configuration) {
         String data = rootNode.get("data").asText();
-        String mimeType = rootNode.get("mime_type").asText();
+        JsonNode media = rootNode.get("media");
```
damianszczepanik: extract `media` into static attribute

src/main/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializer.java
```diff
@@ -24,7 +24,16 @@
...
+        String mimeType;
+
+        if (media != null) {
+            // v3 reports have media object field with type string property
```
damianszczepanik: `v3` is not something I like to have in code as in next month nobody will understand the purpose of such versioning :)

src/main/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializer.java
achingbrain: I'm not sure what you mean by this, but have extracted the logic into an instance method.  It's not a static method because none of the other deserialisers have static helper methods.

src/test/resources/json/cucumber-v2-with-screenshot.json
achingbrain: I've removed the version specific json files and added an embedding with the new media type to the existing `sample.json`.

src/test/resources/json/cucumber-v3-with-screenshot.json
achingbrain: Done.

src/test/java/net/masterthought/cucumber/generators/FeatureReportPageTest.java
achingbrain: In order to have `//given` `//when` `//then` comments as above and to have asserts to appease codacy I've removed this method.

src/test/java/net/masterthought/cucumber/generators/FeatureReportPageTest.java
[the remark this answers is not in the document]
achingbrain: I've made this change.

damianszczepanik:
Thanks!

PR #697, opened 2017-12-18
Reduce memory use when generating pages

sasjo:
The `ReportBuilder` kept a list of pages to generate for no obvious reason. This change removes this list and collects and generates pages in a single pass.

### DIFF
--- a/src/main/java/net/masterthought/cucumber/ReportBuilder.java
+++ b/src/main/java/net/masterthought/cucumber/ReportBuilder.java
@@ -9,7 +9,6 @@ import java.io.OutputStreamWriter;
 import java.io.Reader;
 import java.io.Writer;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
 import java.util.List;
 
 import com.fasterxml.jackson.databind.JsonMappingException;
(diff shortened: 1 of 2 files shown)


sasjo:
Updated with changed method signature. The unit tests on internal structure in `ReportBuilderTest` is the reason why the method even needs to exist and have a return value.

src/main/java/net/masterthought/cucumber/ReportBuilder.java
```diff
@@ -149,34 +149,31 @@ private void copyResources(String resourceLocation, String... resources) {
...
+		numPages += generatePage(new StepsOverviewPage(reportResult, configuration));
+		numPages += generatePage(new FailuresOverviewPage(reportResult, configuration));
+
+		if (configuration.isTrendsStatsFile()) {
+			numPages += generatePage(new TrendsOverviewPage(reportResult, configuration, trends));
+		}
+		return numPages;
+	}
+	
+	private int generatePage(AbstractPage page) {
```
damianszczepanik: now I think, this method has no sense as this is wrapper for the single method.

src/main/java/net/masterthought/cucumber/ReportBuilder.java
```diff
@@ -149,34 +149,31 @@ private void copyResources(String resourceLocation, String... resources) {
...
+		page.generatePage();
+		return 1;
```
damianszczepanik: also method that returns constant value is part of code smell and should hot happen

sasjo:
Restructured code and updated tests to verify using output HTML files instead of return values from private functions. Removed a test that was only testing internal (now removed) code. Code coverage should be unchanged in target file.

damianszczepanik:
Thank you for this contribution!

Issue #682, opened 2017-10-30
OutOfMemory error with large JSON.


manuperera:
Hi,

I have a cucumber JSON file with 1034018 lines (37 MB). The plugin for Jenkins is producing a OutOfMemory error:

```
java.lang.OutOfMemoryError: Java heap space
```
![image](https://user-images.githubusercontent.com/28434493/32166954-b7e2b628-bd67-11e7-8ab6-63fa61af4c92.png)

I have configured Tomcat server where run Jenkins with 8GB of Heap. I don't understand because cucumber report plugin is using all the memory heap.

I appreciate any help you can give me.

Regards,

damianszczepanik:
I see the parsing process has already be started and the failure was caused by velocity which tries to build the page. In that case what I can recommend is to increase Java memroy for Jenkins. 37MB is quite impressive report file

manuperera:
Which is your recommendation aboout de size of memory. 10 GB??, 12 GB??. I would appreciate your opinion.

Regards,

fmarmar:
Hi, I've been debugging this problem cause we are facing the same issue, and we couldn`t solve it increasing the memory.

The real problem is that we are creating and initializing one VelocityEngine instance for html we are going to create, keeping them in memory to later on writing the html to disk. I've been able to make a simply modification and keep the max heap size under 1.5GB.

Probably there is room for improvement, but it's a good improvement. If I get the time I will try to make a pull request. In the meantime, here is the modified file.

[ReportBuilder.zip](https://github.com/damianszczepanik/cucumber-reporting/files/1556675/ReportBuilder.zip)

damianszczepanik:
Will be happy to accept this

PR #698, opened 2017-12-19
Issue #682 performance optimization

fmarmar:
Optimize memory consumption. Extract responsibility from AbstractPage


src/main/java/net/masterthought/cucumber/ReportBuilder.java
```diff
@@ -147,27 +152,38 @@ private void copyResources(String resourceLocation, String... resources) {
...
-
-		new StepsOverviewPage(reportResult, configuration).generatePage();
-		new FailuresOverviewPage(reportResult, configuration).generatePage();
-
-		if (configuration.isTrendsStatsFile()) {
-			new TrendsOverviewPage(reportResult, configuration, trends).generatePage();
-		}
-	}
-	
+    private List<AbstractPage> collectPages(Trends trends) {
```
damianszczepanik: this method has been just refactored because of some memory issue - di you check it with the author?

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -64,58 +34,12 @@ public void generatePage() {
...
+    /**
+     * Populates the given context to merge it with the template
+     * 
+     * @param context
+     */
+    public abstract void preparePageContext(VelocityContext context, Configuration configuration, ReportResult reportResult);
 
-    protected abstract void prepareReport();
-
-    private void generateReport() {
```
damianszczepanik: can you explain this refactor?

src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
```diff
@@ -1,218 +0,0 @@
-package net.masterthought.cucumber.generators;
```
damianszczepanik: class stays, tests are gone?

src/test/java/net/masterthought/cucumber/generators/PageGeneratorTest.java
```diff
@@ -0,0 +1,220 @@
...
+import net.masterthought.cucumber.util.Util;
+
+public class PageGeneratorTest extends PageTest {
+
+	@Rule
+	public ExpectedException thrown = ExpectedException.none();
+
+	@Before
+	public void setUp() {
+		setUpWithJson(SAMPLE_JSON);
```
damianszczepanik: can you change tabs to spaces?

src/test/java/net/masterthought/cucumber/generators/PageGeneratorTest.java
```diff
@@ -0,0 +1,220 @@
...
+		assertThat(context.getChainedContext().getKeys()).hasSize(6);
+
+		Object obj = context.get("counter");
+		assertThat(obj).isInstanceOf(Counter.class);
+		Counter counter = (Counter) obj;
+		assertThat(counter.next()).isEqualTo(1);
+
+	}
+
+}
```
damianszczepanik: missing empty lilne

src/test/java/net/masterthought/cucumber/generators/integrations/PageIntegrationTest.java
```diff
@@ -132,10 +132,10 @@ public void generatePage_onJenkinsConfiguration_generatesSummaryTableWithBuildNu
         configuration.setRunWithJenkins(true);
         configuration.setBuildNumber("123");
 
-        page = new StepsOverviewPage(reportResult, configuration);
+        page = new StepsOverviewPage();
 
         // when
-        page.generatePage();
+        initPageGeneartor().generatePage(page);
```
damianszczepanik: I don't like this convention as there is no clear what is tested here exactly: `initPageGenerator` or `generatePage` method. Will be happy to stay with previous clear convention

src/test/resources/templates/generators/testpage.vm
```diff
@@ -0,0 +1,20 @@
...
+      <div class="message">
+        <pre>Test page</pre>
+      </div>
+      <div>Trying to generate report from following files. Make sure they are valid cucumber report files:</div>
+    </div>
+  </div>
+</div>
+
+</body>
+</html>
```
damianszczepanik: empty line here

src/main/java/net/masterthought/cucumber/ReportBuilder.java
fmarmar: Hi, 

I've seen @sasjo refactored the method using the quick/dirty solution I posted in issue #682 . 

This pull request pretend to be a bettere solution

src/test/java/net/masterthought/cucumber/generators/AbstractPageTest.java
fmarmar: All testeable methods are tested on child classes tests

src/test/java/net/masterthought/cucumber/generators/PageGeneratorTest.java
fmarmar: as you wish

src/test/java/net/masterthought/cucumber/generators/PageGeneratorTest.java
fmarmar: as you wish

src/test/resources/templates/generators/testpage.vm
fmarmar: as you wish

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -32,30 +12,20 @@
...
-    protected final Configuration configuration;
-
-    protected AbstractPage(ReportResult reportResult, String templateFileName, Configuration configuration) {
+    
+    protected AbstractPage(String templateFileName) {
         this.templateFileName = templateFileName;
-        this.reportResult = reportResult;
-        this.configuration = configuration;
-
-        this.engine.init(buildProperties());
```
sasjo: A less intrusive way to reduce memory consumption that I've tried is to just use the singleton `VelocityEngine` API. So far I haven't noticed any negative impact with that change and it only affects the internals of AbstractPage without any other API changes.

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
fmarmar: The idea behind the refactor is to apply the single responsibility principle. An AbstractPage is the link between the velocity template and the necessary data to render the given template. 

The PageGenerator object is the one who initialize the VelocityEngine and using an AbstractPage object render the template with the given data writing it to disk.

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
fmarmar: Velocity singleton model is based in static variables that affects the whole jvm. The use of static variables have a lot of problems, and usually they are the root of memory leaks in j2ee containers.

src/main/java/net/masterthought/cucumber/ReportBuilder.java
sasjo: The patch I submitted was all about being as small as possible to get it into production as quickly as possible. As long as this change doesn't reintroduce the two-step page generation that required all pages to be stored in memory I guess it will only reduce memory consumption even more.

damianszczepanik:
Then my recommendation is to check if already patched version (not yet released) is good enough for performance problem

src/test/java/net/masterthought/cucumber/generators/integrations/PageIntegrationTest.java
fmarmar: Changed

fmarmar:
@damianszczepanik I guess the already patched version is good enough but if you break your leg put it in a splint is also good enough but doctors would recommend put it in a cast.

src/main/java/net/masterthought/cucumber/ReportBuilder.java
```diff
@@ -94,9 +96,8 @@ public Reportable generateReports() {
                 trends = updateAndSaveTrends(reportable);
             }
 
-            // Collect and generate pages in a single pass
             generatePages(trends);
-
+
```
damianszczepanik: please remove such change as they introduce some noise

src/main/java/net/masterthought/cucumber/ReportParser.java
```diff
@@ -55,15 +59,15 @@ public ReportParser(Configuration configuration) {
...
         }
 
-        List<Feature> featureResults = new ArrayList<>();
-        for (int i = 0; i < jsonFiles.size(); i++) {
-            String jsonFile = jsonFiles.get(i);
-            Feature[] features = parseForFeature(jsonFile);
-            LOG.info("File '{}' contain {} features", jsonFile, features.length);
-            setMetadata(features, jsonFile, i);
-            featureResults.addAll(Arrays.asList(features));
+        List<Feature> featureResults = new ArrayList<>(jsonFiles.size() * 10);
```
damianszczepanik: Remove magic number - should not have impact for performance as big as it was worth to provide specified number

src/main/java/net/masterthought/cucumber/ReportParser.java
```diff
@@ -55,15 +59,15 @@ public ReportParser(Configuration configuration) {
...
+        int jsonFileNo = 0;
```
damianszczepanik: prefer previous version as the `jsonFileNo` is accessible only from the loop where it is used

src/main/java/net/masterthought/cucumber/ReportParser.java
```diff
@@ -93,11 +97,15 @@ public ReportParser(Configuration configuration) {
...
-    /** Sets additional information and calculates values which should be calculated during object creation. */
-    private void setMetadata(Feature[] features, String jsonFile, int jsonFileNo) {
-        for (Feature feature : features) {
+    
+    
+    /** 
+     * Collects features, sets additional information and calculates values which should be calculated during object creation. 
+     */
+    private void collectFeaturesAndSetMetadata(Collection<Feature> featuresAccumulator, Feature[] features, String jsonFile, int jsonFileNo) {
+    	for (Feature feature : features) {
```
damianszczepanik: if you remove tab from this line probably there won't be difference as lines before and after are the same

src/main/java/net/masterthought/cucumber/ReportParser.java
```diff
@@ -121,7 +129,10 @@ public void parseClassificationsFiles(List<String> propertiesFiles) {
     private void processClassificationFile(String file) {
         try {
             PropertiesConfiguration config = new PropertiesConfiguration(file);
-            Iterator<String> keys = config.getKeys();
+            
+            @SuppressWarnings("unchecked")
+			Iterator<String> keys = config.getKeys();
```
damianszczepanik: strange formatting or at least tabs should be replaced to spaces

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -32,30 +12,20 @@
...
     /** Name of the HTML file which will be generated. */
     private final String templateFileName;
-    /** Results of the report. */
-    protected final ReportResult reportResult;
-    /** Configuration used for this report execution. */
-    protected final Configuration configuration;
-
-    protected AbstractPage(ReportResult reportResult, String templateFileName, Configuration configuration) {
+    
+    protected AbstractPage(String templateFileName) {
```
damianszczepanik: this changes public API but I see this is the price of this change...

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -32,30 +12,20 @@
...
-        prepareReport();
-        generateReport();
+    
+    /**
+     * Returns HTML file name (with extension) for this report.
+     *
+     * @return HTML file for the report
+     */
+    public String getTemplateName() {
+        return "templates/generators/" + templateFileName;
```
damianszczepanik: can you leave code better than it is now and move this path into private static field?

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -64,58 +34,12 @@ public void generatePage() {
      * @return HTML file for the report
      */
     public abstract String getWebPage();
+    
+    /**
+     * Populates the given context to merge it with the template
+     * 
+     * @param context
```
damianszczepanik: useless information, remove it

src/test/java/net/masterthought/cucumber/ReportBuilderTest.java
```diff
@@ -24,6 +23,7 @@
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
+import mockit.Deencapsulation;
```
damianszczepanik: looks like nothing has changed in this file - if so this is good candidate to be reverted

src/test/java/net/masterthought/cucumber/generators/FeatureReportPageTest.java
```diff
@@ -40,15 +40,15 @@ public void getWebPage_ReturnsFeatureFileName() {
...
 
         // given
+    	VelocityContext context = new VelocityContext();
         Feature feature = features.get(1);
-        page = new FeatureReportPage(reportResult, configuration, feature);
+        page = new FeatureReportPage(feature);
 
         // when
-        page.prepareReport();
+        page.preparePageContext(context, configuration, reportResult);
```
damianszczepanik: this method goes to name of the test (now `prepareReport_`)

src/test/java/net/masterthought/cucumber/generators/PageGeneratorTest.java
```diff
@@ -0,0 +1,252 @@
...
+        // then
+        assertThat(context.getKeys()).hasSize(1);
+        assertThat(context.getChainedContext().getKeys()).hasSize(6);
+
+        Object obj = context.get("counter");
+        assertThat(obj).isInstanceOf(Counter.class);
+        Counter counter = (Counter) obj;
+        assertThat(counter.next()).isEqualTo(1);
+    }
+}
```
damianszczepanik: new line missing

src/test/java/net/masterthought/cucumber/generators/TagsOverviewPageTest.java
```diff
@@ -38,14 +38,14 @@ public void getWebPage_ReturnsTagOverviewReportFileName() {
     public void prepareReport_AddsCustomProperties() {
 
         // given
-        page = new TagsOverviewPage(reportResult, configuration);
+    	VelocityContext context = new VelocityContext();
```
damianszczepanik: here and in other places: this is global or page context as the name does not provide answer for that question

src/test/java/net/masterthought/cucumber/generators/integrations/PageTest.java
```diff
@@ -25,14 +26,19 @@
 public abstract class PageTest extends ReportGenerator {
 
     protected AbstractPage page;
-
+
```
damianszczepanik: trailing space

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
fmarmar: We could keep the previous public API deprecated.

damianszczepanik: Good point, sounds like good compromise.

src/test/java/net/masterthought/cucumber/generators/PageGeneratorTest.java
```diff
@@ -0,0 +1,240 @@
...
+        File reportFile = new File(configuration.getReportDirectory(), ReportBuilder.BASE_DIRECTORY);
+        FileUtils.touch(reportFile);
+
+        // when
+        thrown.expect(ValidationException.class);
+        generatePage(new AbstractPage("test") {
+
+            @Override
+            public void preparePageContext(VelocityContext context, Configuration configuration, ReportResult reportResult) {
+                // TODO Auto-generated method stub
```
damianszczepanik: TODO should be removed or fixed

Issue #675, opened 2017-10-21
Error: no log4j2 configuration file found


serefarikan:
The following error is displayed when using the maven plugin:

> ERROR StatusLogger No log4j2 configuration file found. Using default configuration: logging only errors to the console. Set system property 'org.apache.logging.log4j.simplelog.StatusLogger.level' to TRACE to show Log4j2 internal initialization logging.

The report generation works. The maven project contains only test code and there is a log4j2-test.properties file under resources which also works. I've tried to put log4j2.xml into every possible directory in the project but the error did not go away. 

This integration test project has third party dependencies that use log4j 2 and they're all working fine, processing the log4j2-test.properties file as they should. 

Finally: great plugin, thanks a lot for the work you've done!

Here is my maven pom section related to cucumber-reporting:

><plugin>
                <groupId>net.masterthought</groupId>
                <artifactId>maven-cucumber-reporting</artifactId>
                <version>3.11.0</version>
                <executions>
                    <execution>
                        <id>execution</id>
                        <phase>test</phase>
                        <goals>
                            <goal>generate</goal>
                        </goals>
                        <configuration>
                            <projectName>tstlib-integration-tests</projectName>
                            <outputDirectory>${project.build.directory}/site/cucumber-reports</outputDirectory>
                            <cucumberOutput>${project.build.directory}/cucumber.json</cucumberOutput>
                            <buildNumber>1</buildNumber>
                        </configuration>
                    </execution>
                </executions>
            </plugin>

damianszczepanik:
Does it cause any problem except error message?

serefarikan:
@damianszczepanik  No. Everything seems to work.

KLBonn:
The issue was closed some time ago, but the problem (confusing / misleading / useless ERROR log line) was not solved.

Recently different colleagues of mine asked me about that annoying message, so I thought, maybe a simple log4j2 config file can be included to silence it?

PR #712, opened 2018-03-22
Proposed fix for damianszczepanik/cucumber-reporting#675

KLBonn:
Issue #675 was closed because of missing functional impact.
But that ERROR log message still is there and initially confuses every colleague looking at the output.

A simple fix would be supplying an appropriate LOG4J config file (like you do already for unit tests).


damianszczepanik:
Happy to see this PR - thank you

PR #717, opened 2018-04-12
Use JUL instead of Log4j

aravinder111:
Remove log4j dependency and cleanup to use JUL

Related to: https://github.com/jenkinsci/cucumber-reports-plugin/issues/213

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,9 +13,6 @@
         <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
         <maven.compile.encoding>UTF-8</maven.compile.encoding>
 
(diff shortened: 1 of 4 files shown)


damianszczepanik:
@aravinder111 good job!

aravinder111:
@damianszczepanik Can you tell me when would be the next release tentatively which includes this code change ??

Is there a way for me to build or download the plugin with this change in the meantime(if next release would take a while) ??

Thanks,
Aravind

damianszczepanik:
you can build the plugin via ``mvn package`` or wait a coupe of weeks

PR #718, opened 2018-04-12
Throw exception when embeddingDirectory not present and not created

aravinder111:
Check if embeddingDirectory can be created, otherwise throw ValidationException

sibling PR related to issue https://github.com/jenkinsci/cucumber-reports-plugin/issues/213: https://github.com/damianszczepanik/cucumber-reporting/pull/717

### DIFF
--- a/src/main/java/net/masterthought/cucumber/ReportBuilder.java
+++ b/src/main/java/net/masterthought/cucumber/ReportBuilder.java
@@ -129,7 +129,10 @@ public class ReportBuilder {
     }
 
     private void createEmbeddingsDirectory() {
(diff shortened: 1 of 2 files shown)


damianszczepanik:
When this problem could happen? I'm fine with this change as long as I'm convinced that there is no other way to fix the source of the problem

aravinder111:
@AndyGee Can you provide your input on @damianszczepanik question above ?? 

I haven't observed this personally, but mkdirs() can return false when it fails to create dirs because of permission level. So throwing an Exception when that happens would be useful.

PR #726, opened 2018-05-04
Fix getErrorMessageTitle when string is not splitable.

dennisgranath:
Solves issue #234 ArrayOutOfBoundsException thrown when error_message = "\\n"

### DIFF
--- a/src/main/java/net/masterthought/cucumber/json/Result.java
+++ b/src/main/java/net/masterthought/cucumber/json/Result.java
@@ -37,6 +37,10 @@ public class Result implements Durationable {
     }
 
     public final String getErrorMessageTitle() {
-        return errorMessage.split("[\\r\\n]+")[0];
+        try {
+            return errorMessage.split("[\\r\\n]+")[0];
+        } catch(ArrayIndexOutOfBoundsException e) {
+            return "";
+        }
     }
 }
--- a/src/main/java/net/masterthought/cucumber/json/Result.java
+++ b/src/main/java/net/masterthought/cucumber/json/Result.java
@@ -37,10 +37,6 @@ public class Result implements Durationable {
     }
 
     public final String getErrorMessageTitle() {
-        try {
-            return errorMessage.split("[\\r\\n]+")[0];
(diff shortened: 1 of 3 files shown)


src/main/java/net/masterthought/cucumber/json/Result.java
```diff
@@ -37,6 +37,10 @@ public String getErrorMessage() {
     }
 
     public final String getErrorMessageTitle() {
-        return errorMessage.split("[\\r\\n]+")[0];
+        try {
+            return errorMessage.split("[\\r\\n]+")[0];
+        } catch(ArrayIndexOutOfBoundsException e) {
```
damianszczepanik: definitely not this way - this is more like cure for the result but I prefer code that prevent this situation

src/main/java/net/masterthought/cucumber/json/Result.java
```diff
@@ -37,6 +37,6 @@ public String getErrorMessage() {
     }
 
     public final String getErrorMessageTitle() {
-        return errorMessage.split("[\\r\\n]+")[0];
+        return errorMessage.split("[\\r?\\n]+", -1)[0];
```
damianszczepanik: leave some comments here to describe briefly your intention

src/test/java/net/masterthought/cucumber/json/ResultTest.java
```diff
@@ -77,4 +77,28 @@ public void getErrorMessage_ReturnsErrorMessage() {
...
+        assertThat(errorMessage).isEqualTo("java.lang.AssertionError: ");
+    }
+
+
+    @Test
+    public void getErrorMessageTitle_ReturnsEmptyStringIfNoErrorMessage() {
+
+        Result result = features.get(1).getElements()[0].getSteps()[8].getResult();
+
+        assertThat(result.getErrorMessageTitle()).isEqualTo("");
```
damianszczepanik: if someone leaves \n as the error message I would expect report also displays it as \n instead of empty string - don't you think this would be better?

PR #745, opened 2018-07-25
Fixes ERROR: Step ?Cucumber reports?

jhammell:
Fixes:

ERROR: Step ?Cucumber reports? aborted due to exception:
java.lang.NoSuchMethodError: org.apache.commons.lang3.math.NumberUtils.isCreatable(Ljava/lang/String;)Z
            at net.masterthought.cucumber.generators.AbstractPage.buildGeneralParameters(AbstractPage.java:114)
            at net.masterthought.cucumber.generators.AbstractPage.<init>(AbstractPage.java:52)
            at net.masterthought.cucumber.generators.FeaturesOverviewPage.<init>(FeaturesOverviewPage.java:12)
            at net.masterthought.cucumber.ReportBuilder.generatePages(ReportBuilder.java:151)
            at net.masterthought.cucumber.ReportBuilder.generateReports(ReportBuilder.java:98)
            at net.masterthought.jenkins.CucumberReportPublisher.generateReport(CucumberReportPublisher.java:300)
            at net.masterthought.jenkins.CucumberReportPublisher.perform(CucumberReportPublisher.java:229)
            at hudson.tasks.BuildStepCompatibilityLayer.perform(BuildStepCompatibilityLayer.java:81)
            at hudson.tasks.BuildStepMonitor$1.perform(BuildStepMonitor.java:20)
            at hudson.model.AbstractBuild$AbstractBuildExecution.perform(AbstractBuild.java:744)
            at hudson.model.AbstractBuild$AbstractBuildExecution.performAllBuildSteps(AbstractBuild.java:690)
            at hudson.maven.MavenModuleSetBuild$MavenModuleSetBuildExecution.post2(MavenModuleSetBuild.java:1073)
            at hudson.model.AbstractBuild$AbstractBuildExecution.post(AbstractBuild.java:635)
            at hudson.model.Run.execute(Run.java:1749)
            at hudson.maven.MavenModuleSetBuild.run(MavenModuleSetBuild.java:543)
            at hudson.model.ResourceController.execute(ResourceController.java:97)
            at hudson.model.Executor.run(Executor.java:429)

### DIFF
--- a/src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
+++ b/src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
@@ -11,7 +11,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.math.NumberUtils;
+
 import org.apache.velocity.Template;
 import org.apache.velocity.VelocityContext;
 import org.apache.velocity.app.VelocityEngine;
@@ -109,13 +109,13 @@ public abstract class AbstractPage {
         context.put("build_time", formattedTime);
 
         // build number is not mandatory
-        String buildNumber = configuration.getBuildNumber();
-        if (StringUtils.isNotBlank(buildNumber) && configuration.isRunWithJenkins()) {
-            if (NumberUtils.isCreatable(buildNumber)) {
-                context.put("build_previous_number", Integer.parseInt(buildNumber) - 1);
-            } else {
-                LOG.log(Level.INFO, String.format("Could not parse build number: %1$s.", configuration.getBuildNumber()));
-            }
+        // the build number could contain " ". calling strip will return 'null' to
+        // buildNumber
+        String buildNumber = StringUtils.stripToNull(configuration.getBuildNumber());
+        if (StringUtils.isNumeric(buildNumber) && configuration.isRunWithJenkins()) {
+            context.put("build_previous_number", Integer.parseInt(buildNumber) - 1);
+        } else {
+             LOG.log(Level.INFO, String.format("Could not parse build number: %1$s.", configuration.getBuildNumber()));
         }
     }
 }


src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -109,13 +109,13 @@ private void buildGeneralParameters() {
...
 
         // build number is not mandatory
-        String buildNumber = configuration.getBuildNumber();
-        if (StringUtils.isNotBlank(buildNumber) && configuration.isRunWithJenkins()) {
-            if (NumberUtils.isCreatable(buildNumber)) {
-                context.put("build_previous_number", Integer.parseInt(buildNumber) - 1);
-            } else {
-                LOG.log(Level.INFO, String.format("Could not parse build number: %1$s.", configuration.getBuildNumber()));
-            }
+        // the build number could contain " ". calling strip will return 'null' to
```
damianszczepanik: when this could heppen?

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -109,13 +109,13 @@ private void buildGeneralParameters() {
...
+        // buildNumber
+        String buildNumber = StringUtils.stripToNull(configuration.getBuildNumber());
+        if (StringUtils.isNumeric(buildNumber) && configuration.isRunWithJenkins()) {
+            context.put("build_previous_number", Integer.parseInt(buildNumber) - 1);
+        } else {
+             LOG.log(Level.INFO, String.format("Could not parse build number: %1$s.", configuration.getBuildNumber()));
```
damianszczepanik: what happened here with while spaces do github could not smoothly compare those lines

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
jhammell: After some doing some intense testing I know this condition would never happen.  It was in the original code however and I thought the developer knew something I didn't.  I also did a test where the Jenkins build # was overridden with a semantic version (this does happen where I work) as well as passing a blank string and a string with an empty space.  What learned was that although the BUILD_NUMBER global variable in Jenkins was the value the user defined the number that is passed to this method is always a whole #.  Would you like me to modify the method and submit a new pull request?

I also figured out what the cause of the original failure was in my case.  On the Jenkins master there are some custom jars in java.ext.dirs one of those jars had commons.lang3 as a dependency which inadvertently returned threw the exception.  That dependency wasn't being used in the jar in question, and once it was removed the method works as originally designed.

It's unclear to me as to why the condition throws an exception except to say that it's a CLASSPATH issue.  I am Java 5,6,7 certified but my current role at my place of employment isn't development.

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
jhammell: Keeping the else statement should stay because I didn't test against Jenkins 1.x

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
damianszczepanik: I don't want to keep something what is hard to understand and maintain.

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
damianszczepanik: can you leave this comment in source code so nobody removes this piece some day?

PR #763, opened 2018-10-06
Read and combine all identically named features into one json report

damianszczepanik:
jenkinsci/cucumber-reports-plugin/pull/98

### DIFF
--- a/src/main/java/net/masterthought/cucumber/json/Feature.java
+++ b/src/main/java/net/masterthought/cucumber/json/Feature.java
@@ -1,6 +1,7 @@
 package net.masterthought.cucumber.json;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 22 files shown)


src/main/java/net/masterthought/cucumber/reducers/ReducingMethod.java
```diff
@@ -0,0 +1,18 @@
...
+ * Supported reducing methods.
+ *
+ * @author Damian Szczepanik (damianszczepanik@github)
+ */
+public enum ReducingMethod {
+
+    /**
+     * Merge features if they have same ID so scenrios will be merged into one scenario.
+     */
+    MERGE_FEATURES_BY_ID,
```
damianszczepanik: Currently I support only one reducing/merging method but I can expect more strategies will be added on request

src/main/java/net/masterthought/cucumber/json/support/TagObject.java
```diff
@@ -153,9 +153,4 @@ public Status getStatus() {
     public String getRawStatus() {
         return status.name().toLowerCase();
     }
-
-    @Override
-    public String getDeviceName() {
```
damianszczepanik: I don't care any more about how the report file is named as this may be confused for the report

src/main/java/net/masterthought/cucumber/json/Feature.java
```diff
@@ -24,29 +25,29 @@
     private final String description = null;
     private final String keyword = null;
 
-    private final Element[] elements = new Element[0];
+    private Element[] elements = new Element[0];
     private final Tag[] tags = new Tag[0];
     // End: attributes from JSON file report
 
-    private String jsonFile;
```
damianszczepanik: don't need any more

PR #782, opened 2018-11-11
Fixes #781: NPE with optional step arguments.

### LINKED ISSUE #781: NullPointerException for optional argument
We have a step like the following:
**@Then("^(\\d+) files(?: matching \"(.*)\")? have been failed to be collected for nodes in file (.*)")**

The second parameter is optional. 
It is managed by correctly Cucumber but when it is not provided we get NullPointerException while generating the report.

**JsonReport when the second parameter is not provided:**
`
           "name": "0 files have been failed to be collected for nodes in file BGF/FileCollectionNodes",
            "match": {
              "arguments": [
                {
                  "val": "0",
                  "offset": 0
                },
                {},
                {
                  "val": "BGF/FileCollectionNodes",
                  "offset": 59
                }
              ],
              "location": "FileCollectionResultNotificationSteps.verifyNumOfFilesToBeCollected(int,String,String)"
            },
`

**LOG:**
`
org.apache.velocity.exception.MethodInvocationException: Invocation of method 'format' in  class net.masterthought.cucumber.util.StepNameFormatter threw exception java.lang.NullPointerException at /templates/macros/json/stepName.vm[line 3, column 47]

....
Caused by: java.lang.NullPointerException
	at net.masterthought.cucumber.util.StepNameFormatter.surroundArguments(StepNameFormatter.java:34)
	at net.masterthought.cucumber.util.StepNameFormatter.format(StepNameFormatter.java:19)
`

**JsonReport when the second parameter is provided:**
`
       "name": "2 files matching \"subscriptionType\u003dBSCRECORDINGS\" have been failed to be collected for nodes in file BSC/StatsFileCollectionNodes",
            "match": {
              "arguments": [
                {
                  "val": "2",
                  "offset": 0
                },
                {
                  "val": "subscriptionType\u003dBSCRECORDINGS",
                  "offset": 18
                },
                {
                  "val": "BSC/StatsFileCollectionNodes",
                  "offset": 101
                }
              ],
              "location": "FileCollectionResultNotificationSteps.verifyNumOfFilesToBeCollected(int,String,String)"
            },
`
     
It seems a regression coming with release 4.2.0

ekalin:
When a step has an optional matching argument and the argument is not
matched, the JSON report has an empty object in place of that argument
in the match array. A check has been added to skip those empty
arguments.

### DIFF
--- a/src/main/java/net/masterthought/cucumber/util/StepNameFormatter.java
+++ b/src/main/java/net/masterthought/cucumber/util/StepNameFormatter.java
@@ -31,6 +31,10 @@ public class StepNameFormatter {
 
     private static void surroundArguments(Argument[] arguments, String preArgument, String postArgument, String[] chars) {
         for (Argument argument : arguments) {
+            if (argument.getOffset() == null) {
+                continue;
+            }
+
             int start = argument.getOffset();
             int end = start + argument.getVal().length() - 1;
             chars[start] = preArgument + chars[start];
--- a/src/test/java/net/masterthought/cucumber/util/StepNameFormatterTest.java
+++ b/src/test/java/net/masterthought/cucumber/util/StepNameFormatterTest.java
@@ -79,6 +79,19 @@ public class StepNameFormatterTest extends PageTest {
     }
 
     @Test
+    public void format_optionalArgumentNotMatched() {
+
+        // given
+        Step step = features.get(1).getElements()[0].getSteps()[0];
+
+        // when
+        String formatted = StepNameFormatter.format(step.getName(), step.getMatch().getArguments(), "<arg>", "</arg>");
+
+        // then
+        assertThat(formatted).isEqualTo("the account balance is <arg>100</arg>");
+    }
+
+    @Test
     public void format_shouldEscape() {
 
         // given
--- a/src/test/resources/json/sample.json
+++ b/src/test/resources/json/sample.json
@@ -348,7 +348,8 @@
                                 {
                                     "val": "100",
                                     "offset": 23
-                                }
+                                },
+                                {}
                             ]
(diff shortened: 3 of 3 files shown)


src/main/java/net/masterthought/cucumber/util/StepNameFormatter.java
```diff
@@ -31,6 +31,10 @@ public static String format(String stepName, Argument[] arguments, String preArg
 
     private static void surroundArguments(Argument[] arguments, String preArgument, String postArgument, String[] chars) {
         for (Argument argument : arguments) {
+            if (argument.getOffset() == null) {
+                continue;
+            }
+
             int start = argument.getOffset();
```
damianszczepanik: move ```start``` above and use it for if condition

damianszczepanik: also can we have this case included into sample.json file?

src/main/java/net/masterthought/cucumber/util/StepNameFormatter.java
```diff
@@ -31,6 +31,10 @@ public static String format(String stepName, Argument[] arguments, String preArg
...
             int end = start + argument.getVal().length() - 1;
```
damianszczepanik: keep start and end together

src/main/java/net/masterthought/cucumber/util/StepNameFormatter.java
ekalin: As for moving `start` above, it is possible, but then `start` and `end` would be separated (contrary to what you asked below), because `end` depends on `start` which might be null.

Issue #789, opened 2018-11-21
java.lang.ArrayIndexOutOfBoundsException: 32 at StepNameFormatter.java:40 using 4.2.2 version


biagiolampo:
Moving from 4.1.0 to 4.2.2 for testing 781 I got a different issue on another report

Please refer to attached json and html report.
[Cucumber_HTML_Report.zip](https://github.com/damianszczepanik/cucumber-reporting/files/2603155/Cucumber_HTML_Report.zip)
[SpectrumAnalyzerIT.json.txt](https://github.com/damianszczepanik/cucumber-reporting/files/2612230/SpectrumAnalyzerIT.json.txt)

NOTE: I suggest to improve the report to show what's the affected section of Json file in case of error like this or 781

damianszczepanik:
@ekalin can fix it - JSON file is attached

ekalin:
I've identified the problem and I'm working on a fix.

Can you post the step definition that leads to this step:
```
          {
            "line": 9,
            "name": "Restore the access to the Server",
            "match": {
              "arguments": [
                {
                  "val": "",
                  "offset": 32
                }
              ],
              "location": "CommonSteps.changeServerPermissions(String)"
            },
            "keyword": "And "
          }
```

biagiolampo:
@Given("^(?:A|Restore the access to the) Server *(unreachable|which deny the access|)$")

biagiolampo:
Thanks for the quick resolution 👍

damianszczepanik:
delivered

PR #790, opened 2018-11-24
Highlight step arguments by making them bold. (#771)

damianszczepanik:
- when value for the argument is empty, formatting should be skipped

Issue #789 jenkinsci/cucumber-reports-plugin/issues/280


damianszczepanik:
Does it make sense to add argument wrapper for empty value? Hmmm I'm not sure...

damianszczepanik:
OK, it makes sense, agree

ekalin:
Yeah, checking if the argument is empty is more efficient.

What I would do is a test with the empty argument at the end, and also name the test case in a way that makes it clear that the real problem is when the argument is at the end.

PR #791, opened 2018-11-24
Fixes error when there is an empty match at end of string. Closes #789.

### LINKED ISSUE #789: java.lang.ArrayIndexOutOfBoundsException: 32 at StepNameFormatter.java:40 using 4.2.2 version

ekalin:
When there is an empty match at the end of string, such as
@Given("^the card is valid(.*)$")
(with not extra text after "valid), an ArrayIndexOutOfBounds was
generated.

### DIFF
--- a/src/main/java/net/masterthought/cucumber/util/StepNameFormatter.java
+++ b/src/main/java/net/masterthought/cucumber/util/StepNameFormatter.java
@@ -31,15 +31,27 @@ public class StepNameFormatter {
 
     private static void surroundArguments(Argument[] arguments, String preArgument, String postArgument, String[] chars) {
         for (Argument argument : arguments) {
-            if (argument.getOffset() == null) {
+            if (isNotMatchedArgument(argument)) {
                 continue;
             }
 
             int start = argument.getOffset();
             int end = start + argument.getVal().length() - 1;
+            if (isArgumentAtEndOfString(start, chars)) {
+                continue;
+            }
+
             chars[start] = preArgument + chars[start];
             chars[end] = chars[end] + postArgument;
         }
+    }
+
+    private static boolean isNotMatchedArgument(Argument argument) {
+        return argument.getOffset() == null;
+    }
+
+    private static boolean isArgumentAtEndOfString(int start, String[] chars) {
+        return start >= chars.length;
     }
 
     private static void escape(String[] chars) {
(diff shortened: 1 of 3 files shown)


src/main/java/net/masterthought/cucumber/util/StepNameFormatter.java
```diff
@@ -31,17 +31,29 @@ public static String format(String stepName, Argument[] arguments, String preArg
 
     private static void surroundArguments(Argument[] arguments, String preArgument, String postArgument, String[] chars) {
         for (Argument argument : arguments) {
-            if (argument.getOffset() == null) {
+            if (isNotMatchedArgument(argument)) {
```
damianszczepanik: would rather examine value

src/main/java/net/masterthought/cucumber/util/StepNameFormatter.java
```diff
@@ -31,17 +31,29 @@ public static String format(String stepName, Argument[] arguments, String preArg
...
             chars[end] = chars[end] + postArgument;
         }
     }
 
+    private static boolean isNotMatchedArgument(Argument argument) {
+        return argument.getOffset() == null;
+    }
+
+    private static boolean isArgumentAtEndOfString(int start, String[] chars) {
+        return start >= chars.length;
```
damianszczepanik: what if the val is not empty but should be added at the end of the step?

ekalin: That's testcase format_endsWithArguments.

If the argument is not empty, then the start must be before the end of string.

damianszczepanik: in fact starts = chars.length only if the val is empty :)

PR #792, opened 2018-11-24
Fixes error when there is an empty match at end of string. Closes #789.

### LINKED ISSUE #789: java.lang.ArrayIndexOutOfBoundsException: 32 at StepNameFormatter.java:40 using 4.2.2 version

ekalin:
When there is an empty match at the end of string, such as
@Given("^the card is valid(.*)$")
(with not extra text after "valid), an ArrayIndexOutOfBounds was
generated.

Also avoids wrapping empty arguments with unnecessary tags.

### DIFF
--- a/src/main/java/net/masterthought/cucumber/util/StepNameFormatter.java
+++ b/src/main/java/net/masterthought/cucumber/util/StepNameFormatter.java
@@ -31,7 +31,7 @@ public class StepNameFormatter {
 
     private static void surroundArguments(Argument[] arguments, String preArgument, String postArgument, String[] chars) {
         for (Argument argument : arguments) {
-            if (argument.getOffset() == null) {
+            if (argument.getOffset() == null || StringUtils.isEmpty(argument.getVal())) {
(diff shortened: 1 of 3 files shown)


ekalin:
How about this one? The testcases make it clear that the problem is at the end, but it uses your check based on an empty val, and avoids unnecessary wrapping of empty matches.

ekalin:
This is not necessary anymore since you used the other version.

PR #814, opened 2019-03-12
Fixed NPE when calling getErrorMessageTitle on an Result with `null` …

FloKaemmerer:
Fixed NPE when calling getErrorMessageTitle on an Result with `null` errorMessage

The NPE happens everytime you try to serialize (eg with Jackson ObjectMapper) a passed feature

### DIFF
--- a/src/test/resources/json/sample.json
+++ b/src/test/resources/json/sample.json
@@ -633,6 +633,24 @@
                     }
                 ],
                 "type": "scenario"
+            },
+            {
+                "id": "passed-scenario-with-no-error-message-in-result",
+                "name": "this scenario is passed and does not have a error masse in the result",
+                "keyword": "Scenario",
+                "line": 37,
+                "steps": [
+                    {
+                        "result": {
+                            "duration": 560000,
+                            "status": "passed"
+                        },
+                        "name": "Stream closing",
+                        "keyword": "Given ",
+                        "line": 38
+                    }
+                ],
+                "type": "scenario"
             }
         ],
         "uri": "net/masterthought/example/ATMK.feature"
--- a/src/test/resources/json/sample.json
+++ b/src/test/resources/json/sample.json
@@ -633,24 +633,6 @@
                     }
                 ],
                 "type": "scenario"
-            },
-            {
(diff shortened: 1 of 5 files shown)


FloKaemmerer:
> Add use case to sample.json and test for that part.

Done

src/test/resources/json/sample.json
```diff
@@ -633,6 +633,24 @@
...
+                "name": "this scenario is passed and does not have a error masse in the result",
+                "keyword": "Scenario",
+                "line": 37,
+                "steps": [
+                    {
+                        "result": {
+                            "duration": 560000,
+                            "status": "passed"
+                        },
+                        "name": "Stream closing",
```
damianszczepanik: instead of adding new step without error message can you remove error message from existing step which is not examined so far so there is no need to add mess about modifying existing tests as much as right now?

FloKaemmerer: Thats what I did in my first commit. So I'm not realy sure what you want me to do instead.

I could use the very first feature/step from the sample json, if thats ok with you
`Result result = features.get(0).getElements()[0].getSteps()[0].getResult();`

damianszczepanik: OK, my bad. I misunderstood the first commit now I do. The question is how did you reach such problem as the JSON which is in sample.json and the snipped you have added works fine with current implementation I believe

FloKaemmerer: Generating the html pages indeed works fine.
To give a little context:
I have a Spring boot application that receives cucumber Reports from various sources, processes them using the cucumber-reporting library and stores the generated Pages.
Besides the html pages the application processes the report to gather some more statistic and at some point I tried to serialize your implementation of the feature class. To serialize the class (using jackson) every getter is called. Therefore the NPE is thrown when you try to serialize a passed feature.

damianszczepanik:
Revert then to first commit as it was and I will merge it. Thanks

FloKaemmerer:
> 
> 
> Revert then to first commit as it was and I will merge it. Thanks

Done

PR #822, opened 2019-03-16
Embedding deserialization

FloKaemmerer:
Some implementations of cucumber not always encode Embeddings in Base64.

Cucumber js does not encode plain/text embeddings.

When deserializing an embedding that is not base64 encoded it results in bronken Strings on the report page.

Therefore there is a check needed if the data of an embedding is realy Base64 encoded.
If so, everything is fine and we can create the embedding.
Else we need to base64 encode the data before creation of the embedding.

PS: I'm having a hard time writing a unit test for the check because of the creating of the embedding file which happens in the deserializer.
I would recomend to move the file generation out of the EmbeddingDeserializer (single responsibility) into the report generator.


src/main/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializer.java
```diff
@@ -26,7 +28,13 @@ public Embedding deserialize(JsonNode rootNode, Configuration configuration) {
         String data = rootNode.get("data").asText();
         String mimeType = findMimeType(rootNode);
 
-        Embedding embedding = new Embedding(mimeType, data);
+        Pattern base64Pattern = Pattern.compile("^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$");
```
damianszczepanik: looks like magic number so it I'm not sure if this is proper implementation

src/main/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializer.java
```diff
@@ -26,7 +28,13 @@ public Embedding deserialize(JsonNode rootNode, Configuration configuration) {
...
+        Embedding embedding;
+        if (base64Pattern.matcher(data).matches()) {
+            embedding = new Embedding(mimeType, data);
+        } else {
+            embedding = new Embedding(mimeType, new String(Base64.encodeBase64(data.getBytes(UTF_8)), UTF_8));
```
damianszczepanik: EmbeddingDeserializerTest already has test for this so you can simple copy&paste to cover new case

src/main/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializer.java
FloKaemmerer: moved regex to constant and added comment

src/main/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializer.java
FloKaemmerer: I can't find a EmbeddingDeserializerTest.
You sure the class is UnitTested?

damianszczepanik: Right, you can take a look for other reference https://github.com/damianszczepanik/cucumber-reporting/blob/master/src/test/java/net/masterthought/cucumber/json/deserializers/StatusDeserializerTest.java

src/main/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializer.java
```diff
@@ -14,19 +14,36 @@
...
 public class EmbeddingDeserializer extends CucumberJsonDeserializer<Embedding> {
 
+    /*
+    Base64 Regex matches for the following:
+    ^([A-Za-z0-9+/]{4})* matches group of four chracters from [A-Z, a-z, 0-9, and + /]
+    If a base64 encoded String is not a multiple of four it is padded with '='.
+    (?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$ checks if the String ends with a group of four of the legal characters
+    or it ends with '=' or '=='
+     */
+    private static final String BASE64_MATCHER_REGEX = "^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$";
```
damianszczepanik: Refer to official documentation, API or library so this is clear that this pattern is complete

src/main/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializer.java
FloKaemmerer: I did that.
The Problem I'm facing is the creation of the embedding files on the filesystem.
As far as I can tell you are using Mockito for mocking, with that I'm not able to mock the static call to Files.

src/main/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializer.java
FloKaemmerer: do you mean something like this:
`For details about legal characters and padding refer to https://en.wikipedia.org/wiki/Base64`

damianszczepanik: Nope, rather
https://docs.oracle.com/javase/8/docs/api/java/util/Base64.Decoder.html#decode-java.lang.String-
https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base64.html#isBase64-java.lang.String-

FloKaemmerer: these checks are less specific then the regex.
these checks would consider the String `This is an normal String` as base64 Encoded. Which is obviously wrong.

damianszczepanik: OK, my concern is that I don't want to have own code which decides if this is base64 or not. If that happens I will be happy to merge.

FloKaemmerer: Do you realy rather want a lib to be Included to the Project then using a static RegEx?

damianszczepanik: maybe it is already included or maybe you need to prove you implementation is complete

FloKaemmerer: See added Unite Test

src/main/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializer.java
```diff
@@ -14,32 +14,54 @@
...
 public class EmbeddingDeserializer extends CucumberJsonDeserializer<Embedding> {
 
+    /*
+    Base64 Regex matches for the following:
+    ^([A-Za-z0-9+/]{4})* matches group of four characters from [A-Z, a-z, 0-9, and + /]
+    If a base64 encoded String is not a multiple of four it is padded with '='.
+    (?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$ checks if the String ends with a group of four of the legal characters
+    or it ends with '=' or '=='
+     */
+    private static final String BASE64_MATCHER_REGEX = "^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$";
```
damianszczepanik: and now I found something that solves our problem out of the box https://docs.oracle.com/javase/8/docs/api/java/util/Base64.Decoder.html so try-catch can be used instead of regexp

FloKaemmerer: The decoder is not as strict as the RegEx.
It considers Strings with the wrong length (missing padding with `=`) as legal. Where as the RegEx also checks for the padding.
But for now it will solve our most pressing issues.

src/test/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializerTest.java
```diff
@@ -0,0 +1,72 @@
...
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(value = JsonNode.class)
+public class EmbeddingDeserializerTest {
+
+    private static final String TARGET_DIR = "target" + File.separator;
+    private Configuration configuration;
+
+    @Before
+    public void setUp(){
+        final String directoryPath = TARGET_DIR + "cucumber-html-reports/embeddings";
```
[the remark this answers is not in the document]
FloKaemmerer: changed dir and added @ After for cleanup

src/test/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializerTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.powermock.api.mockito.PowerMockito.mock;
+import static org.powermock.api.mockito.PowerMockito.when;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(value = JsonNode.class)
+public class EmbeddingDeserializerTest {
+
+    private static final String TARGET_DIR = "target" + File.separator;
```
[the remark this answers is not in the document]
FloKaemmerer: changed dir and added @ After for cleanup

src/test/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializerTest.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+        configuration = new Configuration(new File(TARGET_DIR),"TestProject");
+    }
+
+    @Test
+    public void deserialize_OnEncodedData_returnsEmbeddingWithEncodedData(){
+        EmbeddingDeserializer embeddingDeserializer = new EmbeddingDeserializer();
+        JsonNode node = mock(JsonNode.class);
+        String data = "This String will be Base64 encoded";
+        String encodedData = new String(Base64.getEncoder().encode(data.getBytes(StandardCharsets.UTF_8)),StandardCharsets.UTF_8);
```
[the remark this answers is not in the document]
FloKaemmerer: fixed

src/main/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializer.java
damianszczepanik: then please lib, not regexp

src/main/java/net/masterthought/cucumber/json/deserializers/EmbeddingDeserializer.java
```diff
@@ -26,20 +26,35 @@ public Embedding deserialize(JsonNode rootNode, Configuration configuration) {
...
 
         storeEmbedding(embedding, configuration.getEmbeddingDirectory());
 
         return embedding;
     }
 
+    private String getBase64EncodedData(String data) {
+        try{
+            // If we can successfully decode the data we consider it to be base64 encoded,
+            // so we do not need to do anything here
```
damianszczepanik: very helpful documentation - thanks for leaving it here

Issue #841, opened 2019-05-10
Suppressing "empty"/irrelevant hooks [new feature]


KLBonn:
Hi folks,

generally I appreciate seeing output from hooks ( #656 also works like a charm) :)

However with increased use of different hooks for technical purposes, reports are becoming very difficult to read :(
Especially step hooks add a lot of "noise" mostly without carrying relevant information.
(Before and After scenario, Before or After step...) (see attached screenshot
![CucumberReportWithHooks_ScenarioFail_blacked](https://user-images.githubusercontent.com/6554952/57515650-6f46e200-7313-11e9-8590-d1bfca9e3e1e.PNG) )

Even in collapsed state, there can be soo many "Hooks" sections without any information value :(

Me and my colleagues would greatly benefit from an option (toggle param) to suppress any hooks without attachments (just not rendering those lines or even entire "Hooks" sections in HTML pages).

How do you think about the topic? Is it possible, and when/how if so?

damianszczepanik:
Filter by attachment or filter all of them at all?

Can you attach JSON for reference? I wonder by do you have so many hooks

KLBonn:
Thank you for responding  :-)

Ideally: filter by attachment/embedding, both inside scenario and step.
So, show "Hooks" section only if total count of attachments in contained Before/After hook elements is > 0.

See JSON report for reference. 
Sadly I cannot disclose the report corresponding to screenshot above.
So JSON below is from our demo project only -- the second Scenario Before hook (like in screenshot above) is missing here.

[cucumber-json-report.zip](https://github.com/damianszczepanik/cucumber-reporting/files/3177121/cucumber-json-report.zip)

damianszczepanik:
OK, it really makes sense to have possibility to skip such section
```
                        "match": {
                            "location": "ATMScenario.checkBalance(int)"
                        },
```
if there is nothing else than location

damianszczepanik:
What if there is hook with location only and another one with embeddings and are part of the same step/hook?

KLBonn:
Sorry for not responding earlier.
IMO the filter should remove the location-only hook, and keep the one with embeddings, thus also keeping the section containing that second hook (although it would seem there was only one hook).
Does that make sense?

damianszczepanik:
Solution has been prepared, please review https://github.com/damianszczepanik/cucumber-reporting/pull/857
Some code to make it parameterized and tests need to be added before merge

damianszczepanik:
[...]

PR #848, opened 2019-06-07
#841 Suppressing "empty"/irrelevant hooks

RobertoPegoraro:


damianszczepanik:
Have you tried to make it via https://github.com/damianszczepanik/cucumber-reporting/tree/master/src/main/java/net/masterthought/cucumber/json/deserializers and make it customizable like https://github.com/damianszczepanik/cucumber-reporting/blob/master/src/main/java/net/masterthought/cucumber/presentation/PresentationMode.java ?

RobertoPegoraro:
I dont. But I'll try

RobertoPegoraro:
The deserealization is very low level, it is necessary to have access to a level above.

src/main/java/net/masterthought/cucumber/ReportParser.java
```diff
@@ -81,6 +90,31 @@ public ReportParser(Configuration configuration) {
...
+                for (Step step : steps) {
+                    Hook[] after = step.getAfter();
+                    boolean isEmptyHook = false;
+                    for (Hook hook : after) {
+                        if (hook.getEmbeddings().length <= 0 && hook.getOutputs().length <= 0) {
+                            isEmptyHook = true;
+                            break;
+                        }
+                    }
+                    if (isEmptyHook) {
```
damianszczepanik: I don't want to change the model as this may have interrupt other configuration. Instead I recommend to apply this condition in hooks.vm file

damianszczepanik:
Delivered #857

PR #849, opened 2019-06-07
Feature/#841

RobertoPegoraro:
The deserealization is very low level, it is necessary to have access to a level above.


damianszczepanik:
First it must be configurable. Then testable. Finally with reasonable complexity without iterating over all items at the root level rather at the step one.

PR #857, opened 2019-07-10
Skip empty hooks

damianszczepanik:
Issue #841

### DIFF
--- a/src/main/java/net/masterthought/cucumber/json/Hook.java
+++ b/src/main/java/net/masterthought/cucumber/json/Hook.java
@@ -2,6 +2,7 @@ package net.masterthought.cucumber.json;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import org.apache.commons.lang.StringUtils;
 
 import net.masterthought.cucumber.json.deserializers.OutputsDeserializer;
 import net.masterthought.cucumber.json.support.Resultsable;
@@ -37,5 +38,22 @@ public class Hook implements Resultsable {
 
     public Embedding[] getEmbeddings() {
         return embeddings;
+    }
+
+    /**
+     * Checks if the hook has content meaning as it has at least attachment or result with error message.
+     *
+     * @return <code>true</code> if the hook has content otherwise <code>false</code>
+     */
+    public boolean hasContent() {
+        if (embeddings.length > 0) {
+            // assuming that if the embedding exists then it is not empty
+            return true;
+        }
+        if (StringUtils.isNotBlank(result.getErrorMessage())) {
+            return true;
+        }
+        // TODO: hook with 'output' should be treated as empty or not?
+        return false;
     }
 }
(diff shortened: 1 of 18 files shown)


src/main/java/net/masterthought/cucumber/json/Hook.java
```diff
@@ -38,4 +39,21 @@ public Match getMatch() {
...
+     */
+    public boolean hasContent() {
+        if (embeddings.length > 0) {
+            // assuming that if the embedding exists then it is not empty
+            return true;
+        }
+        if (StringUtils.isNotBlank(result.getErrorMessage())) {
+            return true;
+        }
+        // TODO: hook with 'output' should be treated as empty or not?
```
KLBonn: IMO this implementation is OK. I have not seen important output from hooks (aside from embeddings / exception error messages) yet.
If anyone should miss further output from hooks, she can simply disable the hide_empty_hooks feature.
So as far as I am concerned, feel free to remove TODO comment.

src/main/java/net/masterthought/cucumber/util/Util.java
```diff
@@ -76,4 +81,16 @@ public static String toValidFileName(String fileName) {
...
         return Long.toString((long) fileName.hashCode() + Integer.MAX_VALUE);
     }
+
+    /**
+     * Helper method that removes empty hooks from passed array and packs it into new collection.
+     *
+     * @param hooks hooks to be reduced
+     * @return no empty hooks
+     */
+    public static List<Hook> eliminateEmptyHooks(Hook[] hooks) {
```
KLBonn: I am not familiar with the template engine. Could the different types cause problems? You get an array, and return a java.util.List object.
In hooks.vm below you _update_ $hooks with the method result, so you change its type. Do the macros support this?

src/main/resources/templates/macros/json/hooks.vm
```diff
@@ -1,5 +1,9 @@
 #macro(includeHooks $keyword, $hooks, $status, $level)
 
+#if($hide_empty_hooks)
+  #set($hooks = $util.eliminateEmptyHooks($hooks))
+#end
+
 #if(!$hooks.isEmpty())
```
KLBonn: See type-change comment above at Util.java: 
If NOT $hide_empty_hooks, then $hooks still is of type array, isn't it? And array does not have method .isEmpty(), I think?
Or is there some kind of runtime conversion?
Not sure if this is a problem.

src/main/java/net/masterthought/cucumber/util/Util.java
damianszczepanik: it converts smoothly between array and collections

src/main/resources/templates/macros/json/hooks.vm
damianszczepanik: no flag -> stays unchanged as it is right now

isEmpty is the method of velocity template, not pure java in this case

damianszczepanik:
I've tested the change against JSON you provided and hooks were not there at all (except a couple that have some value AFAIR)

KLBonn:
OK, then it looks simply good to me :+1:

PR #862, opened 2019-08-05
Allow undefined and pending steps to not fail

damianszczepanik:
Issue #692

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 [![License](https://img.shields.io/badge/license-GNU%20LGPL%20v2.1-blue.svg)](https://raw.githubusercontent.com/damianszczepanik/cucumber-reporting/master/LICENCE)
 [![Contributors](https://img.shields.io/github/contributors/damianszczepanik/cucumber-reporting.svg)](https://github.com/damianszczepanik/cucumber-reporting/graphs/contributors)
 
-# Publish pretty [cucumber](http://cukes.info/) reports
+# Publish pretty [cucumber](https://cucumber.io/) reports
 
 This is a Java report publisher primarily created to publish cucumber reports on the Jenkins build server.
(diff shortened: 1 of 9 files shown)


damianszczepanik:
With this change user can configure which statuses won't fail scenario and later feature. As presented, even not all steps are passed, scenario remains passed.
skipped

damianszczepanik:
This change introduces most wanted feature that allows to treat undefined/skipped steps as passed and thus not fail scenario. Please review and let me know if this is what you expect!

@a1dutch @SteveDonie @laxersaz @ransom4real @silviatejera @sbpk516 @mpkorstanje @imtiazhossain @plank87 @kth13 @svsforsuccess @maultosar @LucasSq @jess8147 @KLBonn @kalyankp @UltimateGeek @ptrthomas @iragoubi @aamol

imtiazhossain:
Thanks!

mpkorstanje:
In Cucumber this is called strict/no-strict. By default Cucumber run in no-strict mode. Perhaps rather then selecting which statuses are passed you can provide a single strict flag. That way the report andCucumberi use the same terminology.

damianszczepanik:
I haven't use default mode I've been using strict. This is valuable comment and I will try to make it easy to use for everyone who know how strict mode works and for those who like to decide which statuses should fail scenario and which don't.

iragoubi:
Thanks, when will this configuration be available on maven-cucumber-reports ?

damianszczepanik:
The plan stays as: test,  write unit test, release this project and then upgrade Jenkins plugin and others.

SteveDonie:
Excellent! Much excitement from me.

a1dutch:
@damianszczepanik agree about the [strict](https://cucumber.io/docs/cucumber/api/#step-results) flag, this is standard cucumber flag and should be the default behaviour.

Just supporting the strict mode might be a simpler option, cucumber hasn't seen the need to add any additional configuration for what status's are failed.

kasunalwis:
In which version we have this fixed. I trued with 0.9.5.RC1 but didnt work

damianszczepanik:
There is no such version https://github.com/damianszczepanik/cucumber-reporting/releases

PR #867, opened 2019-08-20
Consistence items in carousel

margori:
Improve consistency between elements in items in feature carousel.


damianszczepanik:
Thanks!

PR #870, opened 2019-09-09
Feature: Allow to represent MP4 type video formats in report

ervuks:
This PR will allow embedding MP4 videos inside cucumber report. 

Example of use: 
Execute tests on a mobile device and when test fails, add the video of failed scenario. This allows the possibility to show easier why test failed.


src/main/resources/templates/macros/json/embeddings.vm
```diff
@@ -54,7 +56,26 @@
...
+    <div data-toggle="collapse" data-target="#embedding-$embeddingId" class="collapsable-control">
+      <a>#if($embedding.getName())$embedding.getName()#{else}Attachment $index ($format)#end</a>
+      <a href="embeddings/$embedding.getFileName()" download target="_blank">
+        <span class="download-button glyphicon glyphicon-download-alt"></span>
+      </a>
+    </div>
+    <div id="embedding-$embeddingId" class="collapse collapsable-details">
+      <div class="embedding-content">
+        <video width="854" height="480" controls>
+          <source src="embeddings/$embedding.getFileName()" type="video/mp4">
```
damianszczepanik: I will accept this PR when you add sample video to sample.json so it can be presented in live demo http://damianszczepanik.github.io/cucumber-html-reports/overview-features.html

It can be anything, ideally small. Make sure you meet license once this video becomes part of this project

ervuks: There is already a Base64 data with 1-second video. 
[Original-link](https://youtu.be/Wch3gJG2GJ4)

If this is not acceptable, I will make own video.

damianszczepanik: can't see info about license - recommend to prepare own 'hello world' or similar

will be nice if you add also integration test for video but it is not mandatory

ervuks: I've added a new data sample in `sample.json` with my own video.

damianszczepanik:
Thanks!

PR #877, opened 2019-09-29
expand output and embed contents when EXPAND_ALL_STEPS config is used

### LINKED ISSUE #872: EXPAND_ALL_STEPS doesn't expand output/embed contents
Currently if we use the config EXPAND_ALL_STEPS, only steps gets expanded whereas output/ embed contents doesn't expand. Since user is explicitly specifying this setting, I think it would be ideal to expand output and embed contents too.

Thoughts? I can help with a PR if this is approved.

deepakbenny:
@damianszczepanik - as discussed, this PR adds capability to expand output and embed contents when EXPAND_ALL_STEPS config is used.

resolves #872

### DIFF
--- a/src/main/resources/templates/macros/json/embeddings.vm
+++ b/src/main/resources/templates/macros/json/embeddings.vm
@@ -46,7 +46,7 @@
       <span class="download-button glyphicon glyphicon-download-alt"></span>
     </a>
   </div>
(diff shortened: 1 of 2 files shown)


damianszczepanik:
@deepakbenny thanks!

PR #880, opened 2019-10-11
Added a replaceable merger for features with the same subset of scenarios

dabogee:
Hi! First, I would like to say "Thanks" for all your work on the report plugin because I use it for a long time and really love it! But, right now, I faced with an issue which is described below.

This PR contains an extra logic for merging features with the same subset of scenarios. This situation happens when we need to have results of original tests' run and reruns of the failed tests (of original set of tests) in the same report.

Please, review and let me know if is it OK to be merged?


damianszczepanik:
Give some notes how are you going to merge features. Let me understand your intention before I go through the code

dabogee:
> Give some notes how are you going to merge features. Let me understand your intention before I go through the code

The merging logic of features is based on the following suggestion
* list of features which is passed as a parameter to the ReportResult class is ordered from earlier to latest.

Merging rules:
* Every new feature which is not in the result list is appended to the end. (It already was in the code)
* When the results list already has a feature with such Id then we go down and apply the rules below to the scenarios
	** if there is no scenario with the same Id in the feature then
	- if scenario has a background (which is a previous element in the elements list) then both elements are added to the end of the current feature's elements list.
	- if scenario does not have a backgound then only single element is addede to the elements' list.
	** if there is a scenario with the same Id then:
	- scenario + background case: replace both elements (existing element with Id and its background with new ones)
	- scenario only: replace only given scenario by index in the array.


Note: the latest version of cucumber-json plugin generates start_timestamp field for every scenario. Based on this I can update the element comparator to make sure that we have the newest results in the report.

damianszczepanik:
> ordered from earlier to latest.
How, based on what factor?

> the latest version of cucumber-json plugin generates start_timestamp
but you don't use this field
 
Based on https://github.com/damianszczepanik/cucumber-reporting/pull/880/files#diff-fab832596a584c5349ade611d2196dfdR17 - how did you distinguish whether the scenarios should be merged (parallel run) or  replaced (your implementation)

This is change that has or might have big impact for users so I'm looking for integration tests that proves it works as expected. Copy some items from sample.json into another file, merge and test it.

dabogee:
hi, @damianszczepanik !
I got your point and made the selection more explicit. There is a new MERGE_FEATURES_AND_SCENARIOS_WITH_LATEST reduce method which corresponds to ReportScenarioWithLatestMerger. The previous MERGE_FEATURES_BY_ID implementation stay without any changes and locates in ReportFeatureByIdMerger.

Also I added couple of integration tests here ReportResultMergeTest which show that the merged results looks like the same results were received during a single execution.

Please, take a look when you have a chance!

src/main/java/net/masterthought/cucumber/json/Element.java
```diff
@@ -8,20 +9,27 @@
...
 public class Element implements Durationable {
 
     // Start: attributes from JSON file report
+    private final String id = null;
     private final String name = null;
     private final String type = null;
     private final String description = null;
     private final String keyword = null;
+    private final Integer line = null;
+    @JsonProperty("start_timestamp")
```
damianszczepanik: Add comment that this field is provided by new cucumber version

src/main/java/net/masterthought/cucumber/reducers/ElementComparator.java
```diff
@@ -0,0 +1,26 @@
...
+class ElementComparator implements Comparator<Element> {
+
+    @Override
+    public int compare(Element e1, Element e2) {
+        if (e1 != null && e2 != null && StringUtils.equalsIgnoreCase(e1.getType(), e2.getType())) {
+            if (e1.isScenario()) {
+                return Comparator.nullsFirst(String::compareToIgnoreCase)
+                        .compare(e1.getId(), e2.getId());
+            }
+            return Comparator.nullsFirst(Integer::compare)
```
damianszczepanik: when this can happen, add comment

src/main/java/net/masterthought/cucumber/reducers/ElementComparator.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.Comparator;
+
+/**
+ * Compares scenarios and shows if they have the same Id
+ * or holds on the same line if it's a background.
+ */
+class ElementComparator implements Comparator<Element> {
+
+    @Override
+    public int compare(Element e1, Element e2) {
```
damianszczepanik: element1, element2

src/main/java/net/masterthought/cucumber/reducers/ElementComparator.java
```diff
@@ -0,0 +1,26 @@
...
+        if (e1 != null && e2 != null && StringUtils.equalsIgnoreCase(e1.getType(), e2.getType())) {
+            if (e1.isScenario()) {
+                return Comparator.nullsFirst(String::compareToIgnoreCase)
+                        .compare(e1.getId(), e2.getId());
+            }
+            return Comparator.nullsFirst(Integer::compare)
+                    .compare(e1.getLine(), e2.getLine());
+        }
+        return -1;
```
damianszczepanik: when this can happen?

src/main/java/net/masterthought/cucumber/reducers/ReducingMethod.java
```diff
@@ -13,6 +13,34 @@
...
+     *          scenario only: replace only given scenario by index in the array.
+     *
+     * Example:
+     * Original cucumber report is "cucumber.json". Let's look a situation when couple of tests failed there.
+     * Cucumber runner generates a new report, for example, cucumber-rerun.json as a result of rerun the failed tests.
+     *
+     * In that case you will have a merged report where all failed tests from the original cucumber.json file
+     * are overridden with the results from the cucumber-rerun.json.
+     */
+    MERGE_FEATURES_AND_SCENARIOS_WITH_LATEST,
```
damianszczepanik: MERGE_FEATURES_WITH_RETEST ?

src/main/java/net/masterthought/cucumber/reducers/ReportFeatureMerger.java
```diff
@@ -0,0 +1,11 @@
+package net.masterthought.cucumber.reducers;
+
+import net.masterthought.cucumber.json.Feature;
+
+import java.util.List;
+import java.util.function.Predicate;
+
+public interface ReportFeatureMerger extends Predicate<List<ReducingMethod>> {
+
+    List<Feature> merge(List<Feature> features);
```
damianszczepanik: lack of documentation or at least @see

src/main/java/net/masterthought/cucumber/reducers/ReportScenarioWithLatestMerger.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private void updateElements(Feature feature, Element[] elements) {
+        for (int i = 0; i < elements.length; i++) {
+            Element current = elements[i];
+            if (current.isScenario()) {
+                checkArgument(current.getStartTime() != null, ERROR);
+                int index = find(feature.getElements(), current);
+                boolean hasBackground = isBackground(i - 1, elements);
+
+                if (index < 0) {
```
damianszczepanik: I need some time to review this...

src/main/java/net/masterthought/cucumber/reducers/ReportScenarioWithLatestMerger.java
```diff
@@ -0,0 +1,92 @@
...
+                }
+            }
+        }
+    }
+
+    private boolean replaceIfExists(Element target, Element candidate) {
+        return candidate.getStartTime().compareTo(target.getStartTime()) >= 0;
+    }
+
+    private boolean isBackground(int elementInd, Element[] elements) {
```
damianszczepanik: Ind stands for what?

src/test/java/net/masterthought/cucumber/ReportResultMergeTest.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Test
+    public void parsePartOneTwo_WithFailedRerun() {
+        configuration.addReducingMethod(MERGE_FEATURES_AND_SCENARIOS_WITH_LATEST);
+        setUpWithJson(PART_ONE, PART_TWO, PART_TWO_RERUN_FAILED);
+        assertThat(reportResult.getAllFeatures()).hasSize(3);
+
+        Reportable current = reportResult.getFeatureReport();
+        assertThat(current.getFailedScenarios()).isEqualTo(1);
```
damianszczepanik: .hasSize() ?

src/test/java/net/masterthought/cucumber/reducers/ReportFeatureMergerFactoryTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+import static net.masterthought.cucumber.reducers.ReducingMethod.MERGE_FEATURES_AND_SCENARIOS_WITH_LATEST;
+import static net.masterthought.cucumber.reducers.ReducingMethod.MERGE_FEATURES_BY_ID;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ReportFeatureMergerFactoryTest {
+
+    @Test
+    public void get_NullSafe() {
+        assertThat(new ReportFeatureMergerFactory().get(null)).isInstanceOf(ReportFeatureAppendableMerger.class);
```
damianszczepanik: prefer to  have

// given

// when

// then

sections

src/main/java/net/masterthought/cucumber/reducers/ReducingMethod.java
dabogee: done

src/test/java/net/masterthought/cucumber/ReportResultMergeTest.java
dabogee: no, current.getFailedScenarios() returns an int

src/main/java/net/masterthought/cucumber/json/Element.java
dabogee: done

dabogee:
hi, @damianszczepanik ! Sorry, I don't see any new comments.. could we go forward with this PR?

damianszczepanik:
Currently I'm focusing on commit that was reverted due to some problems. In the meantime you can look at checks which were not passed and https://codecov.io/gh/damianszczepanik/cucumber-reporting/pull/880/diff

Sorry for delay

dabogee:
> Currently I'm focusing on commit that was reverted due to some problems. In the meantime you can look at checks which were not passed and https://codecov.io/gh/damianszczepanik/cucumber-reporting/pull/880/diff
> 
> Sorry for delay

np, thanks for your time! will check the code coverage asap

dabogee:
hi, @damianszczepanik ! I added couple of tests but don't know how to run Codecov Report. Could you start it?

src/main/java/net/masterthought/cucumber/reducers/ReducingMethod.java
```diff
@@ -13,6 +13,34 @@
      */
     MERGE_FEATURES_BY_ID,
 
+    /**
+     * Merge features and scenarios from different JSON files of different runs
+     * into a single report by features' and scenarios' ids.
```
damianszczepanik: I understand the retested features have same ID's?

src/main/java/net/masterthought/cucumber/reducers/ReducingMethod.java
```diff
@@ -13,6 +13,34 @@
...
+     *
+     * Merging rules:
+     * - Every new feature which is not in the result list is appended to the end.
+     *
+     * - When the results list already has a feature with such Id then we go down and apply the rules below to the scenarios:
+     *
+     *      1. if there is no scenario with a given Id in the feature's elements list
+     *         then add the scenario to the end of the list.
```
damianszczepanik: how this is possible to have new scenario if feature is retested?

src/main/java/net/masterthought/cucumber/reducers/ReducingMethod.java
```diff
@@ -13,6 +13,34 @@
...
+     *
+     *      2. if scenario has a background (which is a previous element in the elements list)
```
damianszczepanik: scenario which is in the retested feature or the one which has been already added to merged report?

src/main/java/net/masterthought/cucumber/reducers/ReportFeatureAppendableMerger.java
```diff
@@ -0,0 +1,20 @@
+package net.masterthought.cucumber.reducers;
+
+import net.masterthought.cucumber.json.Feature;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+final class ReportFeatureAppendableMerger implements ReportFeatureMerger {
```
damianszczepanik: short javadoc?

src/main/java/net/masterthought/cucumber/reducers/ReportFeatureByIdMerger.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Override
+    public List<Feature> merge(List<Feature> features) {
+        Map<String, Feature> mergedFeatures = new HashMap<>();
+        for (Feature feature : features) {
+            Feature mergedFeature = mergedFeatures.get(feature.getId());
+            if (mergedFeature == null) {
+                mergedFeatures.put(feature.getId(), feature);
+            } else {
+                mergedFeatures.get(feature.getId()).addElements(feature.getElements());
```
damianszczepanik: would it merge or add?
based on the implementation addElements I guess it adds so after that some elements might be duplicated
https://github.com/damianszczepanik/cucumber-reporting/pull/880/files#diff-8bf6b44089f1dce48374088686c23f67R50

src/main/java/net/masterthought/cucumber/reducers/ReportFeatureByIdMerger.java
```diff
@@ -0,0 +1,30 @@
...
+            }
+        }
+        return new ArrayList<>(mergedFeatures.values());
+    }
+
+    @Override
+    public boolean test(List<ReducingMethod> reducingMethods) {
+        return reducingMethods != null && reducingMethods.contains(ReducingMethod.MERGE_FEATURES_BY_ID);
```
damianszczepanik: null or also empty?

src/main/java/net/masterthought/cucumber/reducers/ReportFeatureMerger.java
```diff
@@ -0,0 +1,20 @@
...
+     * @return list of features which are organized by merger.
+     *
+     * @see net.masterthought.cucumber.reducers.ReportFeatureAppendableMerger
+     * @see net.masterthought.cucumber.reducers.ReportFeatureByIdMerger
+     * @see net.masterthought.cucumber.reducers.ReportFeatureWithRetestMerger
+     *
+     * Merger's type depends on a ReducingMethod which is coming from the configuration.
+     */
+    List<Feature> merge(List<Feature> features);
+}
```
damianszczepanik: worth to redefine https://docs.oracle.com/javase/8/docs/api/java/util/function/Predicate.html#test-T- here with documentation what do you expect from Feature point of view

src/main/java/net/masterthought/cucumber/reducers/ReportFeatureMerger.java
```diff
@@ -0,0 +1,20 @@
+package net.masterthought.cucumber.reducers;
+
+import net.masterthought.cucumber.json.Feature;
+
+import java.util.List;
+import java.util.function.Predicate;
+
+public interface ReportFeatureMerger extends Predicate<List<ReducingMethod>> {
```
damianszczepanik: This change is quite big to review and understand - would you split it into smaller changes? 
The first one would be add this interface with default implementation so after that the report works as now but is ready for further modification

It should really speed up review and merging process

PR #881, opened 2019-10-11
Update velocity to 2.1

embee1981:
Hi there

I've been having some problems with version conflicts between the gradle form of this plugin and owasp gradle plugin as both are using different versions of velocity and I can't seem to get the gradle buildscript to do what I want dependency wise. 

Velocity version is currently at 2.1 (which is what the owasp plugin is using) so I've updated this, removed a redundant depedency on velocity-tools (looks like it was used for the escape tool along while back) and fixed a few deprecation warnings wrt property name changes :-)

The generated demo report looks identical to the hosted one and all tests passed locally.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,7 +74,7 @@
             <id>release-sign-artifacts</id>
             <activation>
                 <property>
-                    <!-- will be set by the release plugin upon performing 
+                    <!-- will be set by the release plugin upon performing
                          mvn release:clean release:prepare release:perform -->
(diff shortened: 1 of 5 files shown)


damianszczepanik:
Thank you!

damianszczepanik:
@embee1981 after this change there is problem with dependencies https://github.com/jenkinsci/cucumber-reports-plugin/pull/331 - would you like to take a look and fix?

PR #883, opened 2019-10-20
Revert "Update velocity to 2.1"

damianszczepanik:
Reverts damianszczepanik/cucumber-reporting#881

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,7 +74,7 @@
             <id>release-sign-artifacts</id>
             <activation>
                 <property>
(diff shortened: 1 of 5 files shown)


damianszczepanik:
@embee1981 I have to revert #881 because there are some problems with Jenkins plugin which I need to consider and fix and till that time I don't want to block new releases.

damianszczepanik:
Changes have been already released as cucumber-reporting-4.11.0 so you can use this release for now

PR #895, opened 2019-11-22
Deactivate resource creation

FloKaemmerer:
Adds Flag to the configuration to deactivate creation of static resources for the generated Report.

The flag defaults to true, so if not configured otherwise the resources will be created alongside the Report.

not creating the static resources is usefull if you use the Generator outside of a Jenkins.

Example usecase:
We have a largenumber of build pipelines that run different Cucumber Testsets.
After each testrun I want to make the results accessible at one central point.
Therefore every Pipeline sends its cucumber-report.json to a central service which then generates the Report and stores it (eg. S3 Bucket).
The Service also provides a UI to access the generated Report.
To minimize the space needed for every Generated Report (they can easily reach 500MB+) I don't want to generate and store the static resources with every Report, but provide them using my service.

currently we filter out the static resources during our upload process.
With the flag we could ommit this filtering logic.


damianszczepanik:
Based on your description it is not clear to me: are you generating report only once or for each JSON?

FloKaemmerer:
I have a large number of buildpipelines that run independet sets of Tests.
Each testrun produces one JSON files (Cucumber-report).
This JSON file is then send to a central Service which generates a html Report for that JSON file.
This HTML report is then stored in a S3 bucket.

so yes, I only generate one Report for each JSON.

damianszczepanik:
Then my advice is to generate report not foe each JSON but for whole set and the problem is fixed with current implementation

FloKaemmerer:
that wont work.

The jsons I get are completly unrelated and need to be in different Reports. 
Furthermore the reports are send to the service at different times, therefore I can't merge them into one and only generate one report.

Imagen a Jenkins pipeline that you run over and over again. In each Jenkins pipeline run you have a set of test that need to be visualized. 
So you have to generate a report for every set independently.

But instead to have the reports generated within the Jenkins, I have a dedicated Service, which generates the reports from the source json.

I have extracted the static resources that are generated with every generation of a report and have placed them as a static resource of my Service. 
Therefore I don't need to generate them over and over again during the generation of the reports.

damianszczepanik:
Understand. However I won't accept this PR as this is to custom to be reused for most cases so I recommend to keep this change in github fork or to redesign your generation process

PR #903, opened 2019-12-07
Build URL and Name

### LINKED ISSUE #902: Build Links
We're using this fantastic plugin with some custom code that allows us to publish the report to Gitlab CI Pages.

Unforatunately there doesn't seem to be a way to configure a free-form build URL, but just a build number, which we cannot use as navigation between different build reports works inherently different there than on Jenkins.

So, therefor the request; Would you be able to add an option to the configuration to provide a free-form URL that points to the build / pipeline that generated the report?

Many thanks!

realdadfish:
Add a globally configurable build name and URL that is rendered on top of every page, in the buildInfo section.

Closes #902

### DIFF
--- a/src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
+++ b/src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
@@ -1,23 +1,5 @@
 package net.masterthought.cucumber.generators;
 
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.OutputStreamWriter;
-import java.io.Writer;
-import java.nio.charset.StandardCharsets;
-import java.util.Properties;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-
-import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.math.NumberUtils;
-import org.apache.velocity.Template;
-import org.apache.velocity.VelocityContext;
-import org.apache.velocity.app.VelocityEngine;
-import org.apache.velocity.app.event.EventCartridge;
-import org.apache.velocity.runtime.resource.loader.ClasspathResourceLoader;
-
 import net.masterthought.cucumber.Configuration;
 import net.masterthought.cucumber.ReportBuilder;
 import net.masterthought.cucumber.ReportResult;
@@ -27,6 +9,19 @@ import net.masterthought.cucumber.reducers.ReducingMethod;
 import net.masterthought.cucumber.util.Counter;
 import net.masterthought.cucumber.util.StepNameFormatter;
 import net.masterthought.cucumber.util.Util;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.math.NumberUtils;
+import org.apache.velocity.Template;
+import org.apache.velocity.VelocityContext;
+import org.apache.velocity.app.VelocityEngine;
+import org.apache.velocity.app.event.EventCartridge;
+import org.apache.velocity.runtime.resource.loader.ClasspathResourceLoader;
+
+import java.io.*;
+import java.nio.charset.StandardCharsets;
+import java.util.Properties;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Delivers common methods for page generation.
@@ -114,10 +109,10 @@ public abstract class AbstractPage {
         String buildUrl = configuration.getBuildUrl();
         if (StringUtils.isNotBlank(buildUrl)) {
             context.put("build_url", buildUrl);
-            String buildName = configuration.getBuildName();
-            if (StringUtils.isBlank(buildName)) {
-                buildName = "Build #" + configuration.getBuildNumber();
-            }
+            String buildName = StringUtils.defaultIfBlank(
+                    configuration.getBuildName(),
+                    "Build #" + configuration.getBuildNumber()
+            );
             context.put("build_name", buildName);
         }
 
--- a/src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
+++ b/src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
@@ -111,6 +111,16 @@ public abstract class AbstractPage {
         context.put("build_project_name", configuration.getProjectName());
         context.put("build_number", configuration.getBuildNumber());
 
+        String buildUrl = configuration.getBuildUrl();
+        if (StringUtils.isNotBlank(buildUrl)) {
+            context.put("build_url", buildUrl);
+            String buildName = configuration.getBuildName();
+            if (StringUtils.isBlank(buildName)) {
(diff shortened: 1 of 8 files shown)


damianszczepanik:
Fix tests first

realdadfish:
Interestingly, mvn test went through locally :(

realdadfish:
Sorry, not used to Maven anymore, missed the clean and therefor missed the test failures. `clean verify` showed them.

realdadfish:
The Travis build failure stems from a misconfiguration I believe. It cannot contact a local Sonarqube instance in the "Static analysis" stage.

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -111,6 +111,16 @@ private void buildGeneralParameters() {
         context.put("build_project_name", configuration.getProjectName());
         context.put("build_number", configuration.getBuildNumber());
 
+        String buildUrl = configuration.getBuildUrl();
+        if (StringUtils.isNotBlank(buildUrl)) {
+            context.put("build_url", buildUrl);
+            String buildName = configuration.getBuildName();
+            if (StringUtils.isBlank(buildName)) {
```
damianszczepanik: https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#defaultIfBlank-T-T-

src/main/resources/templates/macros/page/buildinfo.vm
```diff
@@ -7,6 +7,9 @@
       #if($build_number)
         <th>Number</th>
       #end
+      #if($build_url)
+        <th>Link</th>
```
damianszczepanik: if this is build url why not convert build number as link?

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -111,6 +111,16 @@ private void buildGeneralParameters() {
         context.put("build_project_name", configuration.getProjectName());
         context.put("build_number", configuration.getBuildNumber());
```
damianszczepanik: does it make sense to have build number ...

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -111,6 +111,16 @@ private void buildGeneralParameters() {
...
 
+        String buildUrl = configuration.getBuildUrl();
+        if (StringUtils.isNotBlank(buildUrl)) {
+            context.put("build_url", buildUrl);
+            String buildName = configuration.getBuildName();
+            if (StringUtils.isBlank(buildName)) {
+                buildName = "Build #" + configuration.getBuildNumber();
+            }
+            context.put("build_name", buildName);
```
damianszczepanik: ... and build name at the same time ?

realdadfish: The build number has attached semantics in your case, meaning it must be an integer that is consecutive, so you can determine any previous build link / result link from it in the Jenkins use case.
My use case is however entirely different. First I use Gitlab that has non-consecutive build IDs, secondly the reports cannot be attached to an existing Job / Pipeline directly, but have to be put somewhere else, meaning that a build ID alone without the context of a Job or Pipeline is meaningless.

So, no, both at the same time make no sense, a build number makes most sense in a Jenkins Job environment, while the more generic build name / URL thing I introduced is better for a non-structured build / reporting environment.

If you say you want to get rid / replace the build number by this, fine, I can remove this part, but I did not want to do that before talking to you :)

src/main/resources/templates/macros/page/buildinfo.vm
realdadfish: See above, different semantics.

damianszczepanik:
So I see two options now:
- refactor build name from integer to String with compatibility at Configuration level
- keep two of them together but use only one in template

I'm voting for first solution but I'm happy to hear your concerns

realdadfish:
Well, right now one of the options is only used if the particular one is configured, but yes, I'd also rather vote for option one, but again you'd loose the Jenkins navigation option by this. Clearly if we start with a migration scenario, you can certainly store an int (the build ID) inside the string, but you wouldn't want to fallback and create a Jenkins link from that if the string could be parsed as an int, would you? 

Am 14. Dezember 2019 13:36:49 MEZ schrieb Damian Szczepanik <notifications@github.com>:
>So I see two options now:
>- refactor build name from integer to String with compatibility at
>Configuration level
>- keep two of them together but use only one in template
>
>I'm voting for first solution but I'm happy to hear your concerns
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/damianszczepanik/cucumber-reporting/pull/903#issuecomment-565713419

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

damianszczepanik:
True. I don't see right now the case when we run it via Jenkins plugin and at the same time we'd like to have name instead of build number. Can you?

realdadfish:
> True. I don't see right now the case when we run it via Jenkins plugin and at the same time we'd like to have name instead of build number. Can you?

Exactly, build name only makes sense outside of Jenkins.

> I would say even more: if you have name then having navigation won't have sense...

That as well. And for both things "build_number" is the trigger. If present, it autolinks to previous build numbers, if not, people can manually link their builds. So maybe make it explicit and call it `jenkins_build_number` then?

damianszczepanik:
Ideally some day I may accept PR that make it working with TeamCity but till that time I'm OK to for for it :)

realdadfish:
So, how do we continue here, shall I change anything or are you ok with it as is?

damianszczepanik:
> 
> 
> So, how do we continue here, shall I change anything or are you ok with it as is?

code allows to set name and number - what happens then?

can you attach screenshot please

realdadfish:
> code allows to set name and number - what happens then?

In this case, both are rendered. I don't have a screenshot handy for this as I was preparing this on my work PC and am on holiday now.

damianszczepanik:
OK, so I've tested this change locally. Looks good to me except one thing that started to worry once I've generated the page. when using `PresentationMode.RUN_WITH_JENKINS` it adds link to top menu, when running with other CI it adds to page info section. Can we can another `PresentationMode.RUN_WITH_GITLAB` and locate new URL in place where link to Jenkins page is presented?

If the name of the build is important it can be put into classification

realdadfish:
I'll look into it.

realdadfish:
I think adding another presentation mode does not make much sense, yet alone naming it specific to Gitlab. It just means that users will have to provide an additional setting (the "presentation-mode") in addition to a free-style link.
I'll close this PR for now, because I think the implementation does not lead to anywhere. I'll probably simply post-process the HTML output the plugin generates to include the needed link myself.

PR #904, opened 2019-12-09
Sync version of jackson dependencies up to 2.10.1

dabogee:
Hi, @damianszczepanik !
This is a downgrade of com.fasterxml.jackson.datatype:jackson-datatype-jsr310 to 2.9.10 to avoid an issue with Enforser plugin in the following project https://github.com/jenkinsci/cucumber-reports-plugin

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -209,7 +209,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.9.10.1</version>
+            <version>2.10.1</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.datatype</groupId>


damianszczepanik:
The problem has been already raised https://github.com/jenkinsci/cucumber-reports-plugin/pull/331 but I thought that it was fixed after another upgrade :(
It was some discussion about it https://groups.google.com/forum/?nomobile=true#!searchin/jenkinsci-dev/damian%7Csort:date/jenkinsci-dev/UgQ9wznpybw/9s2Cf3lYCQAJ and as the result I have tried to upgrade parent version https://github.com/jenkinsci/cucumber-reports-plugin/blob/master/pom.xml#L7 and then it worked. Not sure but try with version 3.40

dabogee:
It was fixed for `jackson-databind` dependency, I suppose. But I also added `jackson-datatype-jsr310` in my last PR. And those dependencies have different major versions 2.9.X and 2.10.X.
Right now, I synchronized the major version up to 2.10.X and checked this with 3.54 of `org.jenkins-ci.plugins:plugin`. It works well now.
@damianszczepanik , could you please, release the changes.

PR #894, opened 2019-11-18
Introduced ReportFeatureWithRetestMerger.java

dabogee:
Hi, @damianszczepanik ! This is a next PR with a main part of the changes. It contains the ReportFeatureWithRetestMerger which allows to merge features and scenarios from different JSON files of different runs into a single report by features' and scenarios' ids.
Could you take a look, please?

PR #914, opened 2020-02-05
new ReducingMethod KEEP_ONLY_LATEST_SCENARIO_RUNS

nddipiazza:
add a new reducer that can keep only the latest scenario run.
useful for times when you have flaky scenarios and you do retries on
the given scenario, but don't want the flaked attempts to appear in
report.

### DIFF
--- a/src/main/java/net/masterthought/cucumber/ReportParser.java
+++ b/src/main/java/net/masterthought/cucumber/ReportParser.java
@@ -5,6 +5,9 @@ import com.fasterxml.jackson.databind.InjectableValues;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import com.google.common.collect.Lists;
+
+import net.masterthought.cucumber.json.Element;
 import net.masterthought.cucumber.json.Feature;
 import net.masterthought.cucumber.reducers.ReducingMethod;
 import org.apache.commons.configuration.ConfigurationException;
@@ -22,6 +25,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
+import java.util.ListIterator;
+import java.util.Optional;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -71,6 +76,9 @@ public class ReportParser {
                 continue;
             }
             Feature[] features = parseForFeature(jsonFile);
+            if (configuration.containsReducingMethod(ReducingMethod.KEEP_ONLY_LATEST_SCENARIO_RUNS)) {
+                keepOnlyLatestScenarioRuns(features);
+            }
             LOG.log(Level.INFO, String.format("File '%s' contains %d features", jsonFile, features.length));
             featureResults.addAll(Arrays.asList(features));
         }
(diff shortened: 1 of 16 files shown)


src/main/java/net/masterthought/cucumber/ReportParser.java
```diff
@@ -82,6 +90,42 @@ public ReportParser(Configuration configuration) {
 
         return featureResults;
     }
+    
+    /**
+     * If the JSON file has the same scenarios run multiple times, keep only the
```
damianszczepanik: AFAIR JSON does not have results for multiple times. For that case you would rather have more than one JSON file. Can you explain this case?

src/main/java/net/masterthought/cucumber/ReportParser.java
```diff
@@ -82,6 +90,42 @@ public ReportParser(Configuration configuration) {
...
+                    lastElement = Optional.of(element);
+                    numRemoved = 0;
+                }
+            }
+            addRetryNumberToElmIfNeeded(lastElement, numRemoved);
+            feature.setElements(elementList.toArray(new Element[0]));
+        }
+    }
+
+    private void addRetryNumberToElmIfNeeded(Optional<Element> lastElement, int numRemoved) {
```
damianszczepanik: this is Java, we don't shorten names to `ElmI`

src/main/java/net/masterthought/cucumber/json/Element.java
```diff
@@ -15,7 +15,7 @@
 
     // Start: attributes from JSON file report
     private final String id = null;
-    private final String name = null;
+    private String name = null;
```
damianszczepanik: name is immutable

src/main/java/net/masterthought/cucumber/ReportParser.java
damianszczepanik: also looks similar to https://github.com/damianszczepanik/cucumber-reporting/pull/894/files#diff-858efbefb85a2d5a42faf0d6a15da70bR64

damianszczepanik (changes requested):

ramyasreemaley:
when will this feature be available in plugin?

pawarvaibhav:
This is a super useful update. +1 to get it in as soon as possible.

src/main/java/net/masterthought/cucumber/ReportParser.java
pawarvaibhav: @damianszczepanik - I know this is an old PR. This case happens when we use `TestNGCucumberRunner` and use TestNG's retry analyzer capability to retry tests to disregard flaky tests.

Assuming that we run 1 test with re-run enabled - 

What this does is, in case of flaky behavior, where a test would fail first and then would pass when re-tried - it creates 2 entries in JSON for the same test - 1 for the failed test and 2nd for the same test, re-tried. So the final report shows that there were 2 tests. 

If the re-tried scenario also fails, it is counted as another failed test. so total 2 failed tests
If the re-tried scenario passed, the report says, 1 failed and 1 passed.

This PR seems to solve that problem.

damianszczepanik:
it has conflicts with master branch

nddipiazza:
OK I will get the conflicts fixed and the issues resolved so we can get this merged. ASAP.

src/main/java/net/masterthought/cucumber/ReportParser.java
nddipiazza: Exactly. Sorry I didn't respond earlier, got moved on to other priorities and never had a chance to come back and answer the questions/fixes.

src/main/java/net/masterthought/cucumber/json/Element.java
nddipiazza: i change this to mutable because i use the Name in the report because I append " (${numRetries)" so that it shows up in the report. 

alternatively we could create a new element in the html/xslt or whatever.

stevegome:
Hi Team, can anyone please resolve the conflicts for the ^ fix, I am facing a similar issue whereing
- Using ReTry Analyzer of TestNG which re-reruns the failed testcase
- However the cucumber.json file shows multiple status for the same scenario
- My requirement is that it should only update result for the last run

I would really appreciate if someone can point me in the right direction if the issue has already been fixed.

stevegome:
@damianszczepanik Greetings. Can someone please resolve the merge conflicts and get this PR merged please ?

JithinPillai1592:
Hi Team,
I would really appreciate if you could resolve the conflicts and merge the fix.
The duplicate status for the same retried test is confusing to share with other members

JithinPillai1592:
I seeing the behaviour described by @pawarvaibhav 
But when I try with the cucumber re-run plugin instead IRetryAnalyzer the result gets combined

PR #919, opened 2020-02-12
Added optional directory qualifier/suffix to configuration and use it with ReportBuilder.BASE_DIRECTORY

shchukax:
This library is used by jenkinsci/cucumber-reports-plugin. That plugin currently only supports one report per job run.  While trying to extend it to allow multiple reports, my work was limited by the fact that this library has hardcoded destination directory for the report.

To facilitate changes in jenkinsci/cucumber-reports-plugin, I'm adding an optional directoryQualifier parameter to the Configuration, which is then used as a suffix for the reports' target directory.  Now, instead of `ReportBuilder.BASE_DIRECTORY` (which is always `cucumber-html-reports`), the target would be `ReportBuilder.BASE_DIRECTORY + configuration.getDirectoryQualifier` (which would be `cucumber-html-reports<specified-suffix>`). This qualifier/suffix is optional. If it's not specified, the behaviour of the library is unchanged.


shchukax:
Is travic-ci build broken?  It fails with an error that it can't connect to soniqcube.

src/main/java/net/masterthought/cucumber/Configuration.java
```diff
@@ -136,13 +138,31 @@ public String getProjectName() {
         return projectName;
     }
 
+    /**
+     * Sets directory qualifier.
+     *
+     * @param directoryQualifier directory qualifier
+     */
+    public void setDirectoryQualifier(String directoryQualifier) {
```
damianszczepanik: was it intentional to use similar name to https://github.com/damianszczepanik/cucumber-reporting/commit/3b3bef7a8d32841310587629e607bab7c1a9af75#diff-8bf6b44089f1dce48374088686c23f67R78 ?

src/main/java/net/masterthought/cucumber/Configuration.java
```diff
@@ -136,13 +138,31 @@ public String getProjectName() {
...
+    }
+
     /**
      * Gets directory where the attachments are stored.
      *
      * @return directory for attachment
      */
     public File getEmbeddingDirectory() {
-        return new File(getReportDirectory().getAbsolutePath(), ReportBuilder.BASE_DIRECTORY
+        return new File(getReportDirectory().getAbsolutePath(), ReportBuilder.BASE_DIRECTORY + this.getDirectoryQualifier()
```
damianszczepanik: looks like directory separator is missing here

src/main/java/net/masterthought/cucumber/ReportBuilder.java
```diff
@@ -135,7 +135,7 @@ private void createEmbeddingsDirectory() {
     private void copyResources(String resourceLocation, String... resources) {
         for (String resource : resources) {
             File tempFile = new File(configuration.getReportDirectory().getAbsoluteFile(),
-                    BASE_DIRECTORY + File.separatorChar + resourceLocation + File.separatorChar + resource);
+                    BASE_DIRECTORY + configuration.getDirectoryQualifier() + File.separatorChar + resourceLocation + File.separatorChar + resource);
```
damianszczepanik: and here

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
```diff
@@ -70,7 +70,7 @@ private void generateReport() {
 
         Template template = engine.getTemplate("templates/generators/" + templateFileName);
         File reportFile = new File(configuration.getReportDirectory(),
-                ReportBuilder.BASE_DIRECTORY + File.separatorChar + getWebPage());
+                ReportBuilder.BASE_DIRECTORY + configuration.getDirectoryQualifier() + File.separatorChar + getWebPage());
```
damianszczepanik: and here :)

src/main/resources/templates/macros/page/navigation.vm
```diff
@@ -7,9 +7,9 @@
         #if ($run_with_jenkins)
           <li role="presentation"><a href="../">Jenkins</a></li>
           #if ($build_previous_number)
-            <li role="presentation"><a href="../../$build_previous_number/cucumber-html-reports/$report_file">Previous results</a></li>
+            <li role="presentation"><a href="../../$build_previous_number/cucumber-html-reports$directory_qualifier/$report_file">Previous results</a></li>
```
damianszczepanik: did you address my concern from last review about compatibility?

src/main/java/net/masterthought/cucumber/Configuration.java
shchukax: Nope, just happened to be this.  I'm actually thinking about changing it to `directorySuffix` - let me know if it's preferred.

src/main/java/net/masterthought/cucumber/Configuration.java
shchukax: No, it's not missing; it's not needed - as this is just a suffix to the current directory name.  It can be passed as `_abcd` for example.

src/main/java/net/masterthought/cucumber/generators/AbstractPage.java
shchukax: No, it's not missing; it's not needed - as this is just a suffix to the current directory name. It can be passed as _abcd for example.

src/main/java/net/masterthought/cucumber/ReportBuilder.java
shchukax: No, it's not missing; it's not needed - as this is just a suffix to the current directory name. It can be passed as _abcd for example.

src/main/resources/templates/macros/page/navigation.vm
shchukax: Sorry, I'm not sure which one you are referring to. If this is about navigation to previous/latest, then this particular change is exactly for that.

damianszczepanik: Previous link does not work after this change

src/main/java/net/masterthought/cucumber/Configuration.java
damianszczepanik: Then recommend to change it to avoid confusing

shchukax:
I renamed `directoryQualifier` to `directorySuffix` and `directory_qualifier` to `directory_suffix` everywhere. As for the link to the previous report not working - it does work as long as the previous report was done with the same suffix - that's the whole idea. If I am generating multiple reports with different directory suffixes, each one will navigate to its own previous one.

src/main/java/net/masterthought/cucumber/Configuration.java
```diff
@@ -136,13 +138,31 @@ public String getProjectName() {
...
+    }
+
     /**
      * Gets directory where the attachments are stored.
      *
      * @return directory for attachment
      */
     public File getEmbeddingDirectory() {
-        return new File(getReportDirectory().getAbsolutePath(), ReportBuilder.BASE_DIRECTORY
+        return new File(getReportDirectory().getAbsolutePath(), ReportBuilder.BASE_DIRECTORY + this.getdirectorySuffix()
```
damianszczepanik: IMO it would be nice to have some separator such as underscore between BASE_DIRECTORY and suffix. Consider this

src/main/java/net/masterthought/cucumber/Configuration.java
```diff
@@ -136,13 +138,31 @@ public String getProjectName() {
...
+    public void setdirectorySuffix(String directorySuffix) {
+        this.directorySuffix = directorySuffix;
+    }
+
+    /**
+     * Returns directory qualifier
+     *
+     * @return directory qualifier
+     */
+    public String getdirectorySuffix() {
```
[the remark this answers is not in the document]
shchukax: Sorry, a typo during previous renaming exercise - fixed now (both in setter and getter)

src/main/java/net/masterthought/cucumber/Configuration.java
shchukax: I'd rather let the end user decide whether such separator is needed - if yes, then it can be passed as part of the suffix (as I do myself in my own compiled version).

src/main/java/net/masterthought/cucumber/Configuration.java
```diff
@@ -136,13 +138,31 @@ public String getProjectName() {
...
+    }
+
     /**
      * Gets directory where the attachments are stored.
      *
      * @return directory for attachment
      */
     public File getEmbeddingDirectory() {
-        return new File(getReportDirectory().getAbsolutePath(), ReportBuilder.BASE_DIRECTORY
+        return new File(getReportDirectory().getAbsolutePath(), ReportBuilder.BASE_DIRECTORY + this.getDirectorySuffix()
```
damianszczepanik: if you look how currently BASE_DIRECTORY looks like then IMO its better to have separator _ or / between - otherwise it might look strange
https://github.com/damianszczepanik/cucumber-reporting/pull/919/files#diff-e17c4a370f9a9da4a1fb9017744ebba7L46

shchukax: Ok, I made the change now to use `_` as suffix separator.

src/main/java/net/masterthought/cucumber/Configuration.java
```diff
@@ -136,14 +139,41 @@ public String getProjectName() {
...
+        return StringUtils.defaultString(directorySuffix);
+    }
+
+    /**
+     * Returns directory suffix with separator prepended if necessary
+     *
+     * @return directory suffix with prepended separator
+     */
+    public String getDirectorySuffixWithSeparator() {
+        return StringUtils.isEmpty(directorySuffix) ? "" : SUFFIX_SEPARATOR + directorySuffix;
```
damianszczepanik: can't find unit test for new methods

src/main/resources/templates/macros/page/navigation.vm
```diff
@@ -7,9 +7,9 @@
         #if ($run_with_jenkins)
           <li role="presentation"><a href="../">Jenkins</a></li>
           #if ($build_previous_number)
-            <li role="presentation"><a href="../../$build_previous_number/cucumber-html-reports/$report_file">Previous results</a></li>
+            <li role="presentation"><a href="../../$build_previous_number/cucumber-html-reports$directory_suffix/$report_file">Previous results</a></li>
```
damianszczepanik: looks good - did you check if old configuration with new plugin works fine? can you move back and forward between older and newer reports?

src/main/java/net/masterthought/cucumber/Configuration.java
shchukax: Ah! good point - missed that. I added the unit tests now.

src/main/resources/templates/macros/page/navigation.vm
shchukax: When using the new version, if the suffix is not set or is set to an empty string, then navigation to previous reports work whether generated with the old or the new version. If a suffix is set, then there will not be "previous" report - so there won't be a link.  Previous/next is based on the same suffix.

damianszczepanik:
tests are still missing

shchukax:
> tests are still missing

D'oh! Committed but didn't push last time - all in now.

shchukax:
D'oh! Committed but forgot to push last time - all in now.

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
```diff
@@ -153,6 +150,51 @@ public void getTagsToExcludeFromChart_ReturnsEmptyList() {
...
+        String directorySuffix = "test";
+        Configuration configuration = new Configuration(outputDirectory, projectName);
+
+        // when
+        configuration.setDirectorySuffix(directorySuffix);
+
+        // then
+        Class<Configuration> configClass = Configuration.class;
+        try {
+            Field suffixSeparator = configClass.getDeclaredField("SUFFIX_SEPARATOR");
```
damianszczepanik: I don't get it why this must be so complicated :)

shchukax: I didn't want to hardcode the separator - so had to get the value of a private variable.  Anyway, I changed it now to have a hardcoded '_' separator in the test.

src/test/java/net/masterthought/cucumber/ConfigurationTest.java
```diff
@@ -153,6 +150,51 @@ public void getTagsToExcludeFromChart_ReturnsEmptyList() {
...
+            suffixSeparator.setAccessible(true);
+            assertThat(configuration.getDirectorySuffixWithSeparator()).isEqualTo(suffixSeparator.get(null) + directorySuffix);
+        } catch (NoSuchFieldException|IllegalAccessException ex) {
```
damianszczepanik, line 183: still not changed

shchukax: Argh! Another one "commit-no push".  Done now.

damianszczepanik:
@shchukax well done!

shchukax:
Thanks!

Once this is built and updated in Maven Central, I'll get back to the corresponding jenkins plugin updates.
-Aleks

PR #923, opened 2020-04-12
Cucumber reporting - New Generation UI

fslev:
It is quite too early for this pull request to be approved, since it contains code that still needs some cleanup. It doesn't need to be approved at all: it's just a UI demo.
The idea behind is that I want you to take a look on the new revamped UI version of cucumber-reporting. Basically, I've tried to make it cleaner (Each passed step has now a white background. Having other background color it can distract you from the text within). Also added some contrast style where important sections are displayed, thus making the UI also more memorable. 
You can take a short look by running the LiveDemoTest


damianszczepanik:
Nice, probably more modern but less compact what IMO is disadvantage

fslev:
Indeed. Some compact related styles were sacrificed. Compact is good for small screens. However, in a demo session, on a projector, it's quite hard to follow all the details. Pros and cons. Thus a compact feature toggle could be in place at some point in time.

Issue #925, opened 2020-04-15
Feature Request: Disable or speed up "Tags" page


mattventura:
Due to their size, our reports often take a while to generate (15+ seconds, sometimes as much as 40 seconds). I did some investigation and found that the majority of this time was spent in `ReportBuilder.generatePages` in this loop:

```
            for (TagObject tagObject : reportResult.getAllTags()) {
                new TagReportPage(reportResult, configuration, tagObject).generatePage();
            }
```

Since I don't use these pages anyway, I'd like to have an option to disable this completely. Using `configuration.setTagsToExcludeFromChart(".*")` resulted in an empty tags page, but did not improve performance noticeably. Commenting out the loop entirely was the only thing that had a significant performance gain.

Alternatively, if something could be done to improve the performance of the tags pages significantly, that would also be welcome.

damianszczepanik:
@mattventura one more test before release? I fixed issue you have reported which was related to missing page.

mattventura:
@damianszczepanik looks good to me.

PR #927, opened 2020-04-17
Disable or speed up "Tags" page

damianszczepanik:
Issue #925


damianszczepanik:
@mattventura this change speeds up tags generation however I'm not sure if this is enough for you. Can you test this change? Or share JSON file so I can examine fix

mattventura:
Yep, this speeds it up significantly if I exclude all the tags. Unfortunately I can't provide my test file due to company info, but it cut the time down significantly. Thank you!

damianszczepanik:
Then is it good enough to make this change ready for release? If so I will update tests and merge it.

mattventura:
The only thing I noticed is that the excluded tags still show up in the tags overview page, but they link to a nonexistent page. For usability, it might be nice to not have them be normal text rather than links if the tag was filtered out. Other than that, looks fine to me.

PR #932, opened 2020-04-27
Make SUFFIX_SEPARATOR static variable public

shchukax:
SUFFIX_SEPARATOR static variable (introduced by me in previous PR) was private, but it makes more sense for it to be PUBLIC so that it could be used in downstream projects (e.g. cucumber-reports-plugin for jenkins) to improve consistency.  Also for consistency, it is moved to ReportBuilder class, next to BASE_REPORT_DIR, as it is used along with it to construct the location where the report should be placed.


src/main/java/net/masterthought/cucumber/ReportBuilder.java
```diff
@@ -43,6 +43,11 @@
      * Subdirectory where the report will be created.
      */
     public static final String BASE_DIRECTORY = "cucumber-html-reports";
+    /**
+     * Separator between main directory name and specified suffix
+     */
+    public static final String SUFFIX_SEPARATOR = "_";
```
damianszczepanik, line 49: reduntant with https://github.com/damianszczepanik/cucumber-reporting/blob/master/src/main/java/net/masterthought/cucumber/Configuration.java#L23

shchukax: I must have missed during merges.  Variable PREFIX_SEPARATOR in Configuration class is no longer used - and I pushed another commit to remove it.

shchukax: I must have missed something during merges.  Variable PREFIX_SEPARATOR in Configuration class is not used any more - and I pushed another commit removing it.

damianszczepanik:
@shchukax released!

PR #959, opened 2020-08-13
Enable generation of Reports from a List of Features

FlorianKaemmerer:
this is just a rough idea how you got enable generation of the Reports from a List of Features

### DIFF
--- a/src/main/java/net/masterthought/cucumber/ReportBuilder.java
+++ b/src/main/java/net/masterthought/cucumber/ReportBuilder.java
@@ -1,33 +1,25 @@
 package net.masterthought.cucumber;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.OutputStreamWriter;
-import java.io.Reader;
-import java.io.Writer;
-import java.nio.charset.StandardCharsets;
-import java.util.List;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectWriter;
 import com.fasterxml.jackson.databind.SerializationFeature;
-import net.masterthought.cucumber.generators.ErrorPage;
-import net.masterthought.cucumber.generators.FailuresOverviewPage;
-import net.masterthought.cucumber.generators.FeatureReportPage;
-import net.masterthought.cucumber.generators.FeaturesOverviewPage;
-import net.masterthought.cucumber.generators.StepsOverviewPage;
-import net.masterthought.cucumber.generators.TagReportPage;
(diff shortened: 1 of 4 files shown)


src/main/java/net/masterthought/cucumber/ReportBuilder.java
```diff
@@ -64,45 +56,46 @@
...
+        return generateReportsFromFiles(jsonFiles);
+    }
 
+    /**
+     * Parses provided files and generates the report. When generating process fails
+     * report with information about error is provided.
+     *
+     * @return stats for the generated report
+     */
+    public Reportable generateReportsFromFiles(List<String> pathsTojsonFiles) {
```
damianszczepanik, line 88: this changes the API - will you update also maven and jenkins plugin which use this library as dependency?

src/main/java/net/masterthought/cucumber/ReportBuilder.java
```diff
@@ -121,6 +114,62 @@ public Reportable generateReports() {
...
+
+        // Collect and generate pages in a single pass
+        generatePages(trends);
+
+        return reportable;
+    }
+
+    private void generateEmbeddings() {
+        // create directory for embeddings before files are generated
+        createEmbeddingsDirectory();
```
damianszczepanik, line 143: I think you have already removed this method by second PR?

src/main/java/net/masterthought/cucumber/ReportBuilder.java
```diff
@@ -121,6 +114,62 @@ public Reportable generateReports() {
...
+
+        reportResult.getAllFeatures().forEach(feature -> {
+            for (Element element : feature.getElements()) {
+                for (Hook hook : element.getAfter()) {
```
damianszczepanik, line 147: after....

src/main/java/net/masterthought/cucumber/ReportBuilder.java
```diff
@@ -121,6 +114,62 @@ public Reportable generateReports() {
...
+                    storeEmbedding(hook.getEmbeddings(), configuration.getEmbeddingDirectory());
+                }
+                for (Hook hook : element.getBefore()) {
```
damianszczepanik, line 150: ... should be after before :)

src/test/java/net/masterthought/cucumber/generators/integrations/FeatureReportPageIntegrationTest.java
```diff
@@ -425,8 +405,8 @@ private static void validateHook(HookAssertion[] hookAssertions, Hook[] hooks, S
         }
     }
 
-    private void assertEmbeddingFileExist(Embedding embedding) {
-        File file = new File(configuration.getEmbeddingDirectory(), embedding.getFileName());
-        assertThat(file).exists();
-    }
+//    private void assertEmbeddingFileExist(Embedding embedding) {
```
damianszczepanik, line 408: unused method should be removed, not commented out

git keeps history of changes

damianszczepanik (changes requested):

Issue #969, opened 2020-10-31
Improve Extensions for Report Attachement


ghostcity:
Dear Damian, in many cases the HTML report does not specify a correct attachment 's extension, even if it could be done.

For example, Mime-Types with additional encoding (e.g., "text/html; charset=UTF-8"), the extension is "unknown" instead of "html". Without encoding, "html" is used.

For further example, the subtype of an unknown Mime-Type (e.g., "text/php") might be a good hint to deduce the extension.

Even further, the name itself could be interpreted as a file-name and, thus, the given extension could be used.

damianszczepanik:
Prepare PR with changes so we can review then and discuss before merge

PR #971, opened 2020-10-31
969 improve extensions for report attachment

ghostcity:
Dear Damian, I read your [contributing guidlines](https://github.com/damianszczepanik/cucumber-reporting/blob/master/CONTRIBUTING.md) and would like to offer an implementation of the issue #969 (Improve Extensions for Report Attachment).

My first pull request did not test all of the decision branches of the new code. Now, it does.
Please let me know if I shall improve anything else of that solution.

Thank you in advance, Stefan.

### DIFF
--- a/src/main/java/net/masterthought/cucumber/json/Embedding.java
+++ b/src/main/java/net/masterthought/cucumber/json/Embedding.java
@@ -126,16 +126,20 @@ public class Embedding {
                 return "xls";
             default:
                 // assert the name is file-name formatted --> try file-name extension
-                if (name != null && name.contains(".")) {
+                if ((name != null) && name.contains(".")) {
                     String extension = name.substring(name.lastIndexOf('.') + 1);
                     // the extension might by usable
-                    if (extension.matches(FILE_EXTENSION_PATTERN)) return extension;
+                    if (extension.matches(FILE_EXTENSION_PATTERN)) {
+                        return extension;
+                    }
                 }
                 // assert the mime-type contains a subtype --> try subtype
                 if (mime.contains("/")) {
                     String subtype = mime.substring(mime.indexOf('/') + 1);
                     // the subtype might by usable
-                    if (subtype.matches(FILE_EXTENSION_PATTERN)) return subtype;
+                    if (subtype.matches(FILE_EXTENSION_PATTERN)) {
+                        return subtype;
+                    }
                 }
                 // if nothing works the extension is unknown
                 return UNKNOWN_FILE_EXTENSION;
--- a/src/main/java/net/masterthought/cucumber/json/Embedding.java
+++ b/src/main/java/net/masterthought/cucumber/json/Embedding.java
@@ -104,6 +104,19 @@ public class Embedding {
             case "application/vnd.ms-excel":
                 return "xls";
             default:
+                // assert the name is file-name formatted --> try file-name extension
+                if (name != null && name.contains(".")) {
+                    String extension = name.substring(name.lastIndexOf('.') + 1);
+                    // the extension might by usable
+                    if (extension.matches("[a-z0-9]+")) return extension;
+                }
+                // assert the mime-type contains a subtype --> try subtype
+                if (mime.contains("/")) {
+                    String subtype = mime.substring(mime.indexOf('/') + 1);
+                    // the subtype might by usable
+                    if (subtype.matches("[a-z0-9]+")) return subtype;
+                }
+                // if nothing works the extension is unknown
                 return "unknown";
         }
     }
--- a/src/main/java/net/masterthought/cucumber/json/Embedding.java
+++ b/src/main/java/net/masterthought/cucumber/json/Embedding.java
@@ -1,6 +1,7 @@
 package net.masterthought.cucumber.json;
 
 import java.nio.charset.StandardCharsets;
+import java.util.Locale;
 
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import org.codehaus.plexus.util.Base64;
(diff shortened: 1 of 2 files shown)


src/main/java/net/masterthought/cucumber/json/Embedding.java
```diff
@@ -63,28 +64,35 @@ public String getFileId() {
     }
 
     public String getExtension() {
-        switch (mimeType) {
+        // prepare/ensure switch-case matching
```
damianszczepanik, line 67: some reference to official documentation is more than welcome

src/main/java/net/masterthought/cucumber/json/Embedding.java
```diff
@@ -96,6 +104,19 @@ public String getExtension() {
             case "application/vnd.ms-excel":
                 return "xls";
             default:
+                // assert the name is file-name formatted --> try file-name extension
+                if (name != null && name.contains(".")) {
+                    String extension = name.substring(name.lastIndexOf('.') + 1);
+                    // the extension might by usable
+                    if (extension.matches("[a-z0-9]+")) return extension;
```
damianszczepanik, line 111: would it match "www.google.com/gmail ?

src/main/java/net/masterthought/cucumber/json/Embedding.java
```diff
@@ -96,6 +104,19 @@ public String getExtension() {
...
+                }
+                // assert the mime-type contains a subtype --> try subtype
+                if (mime.contains("/")) {
+                    String subtype = mime.substring(mime.indexOf('/') + 1);
+                    // the subtype might by usable
+                    if (subtype.matches("[a-z0-9]+")) return subtype;
```
damianszczepanik, line 117: format with {}

src/test/java/net/masterthought/cucumber/json/EmbeddingTest.java
```diff
@@ -90,6 +90,19 @@ public void getExtension__OnCommonMimeType_ReturnsFileExtension() {
         assertThat(extension).isEqualTo("html");
     }
 
+    @Test
+    public void getExtension__OnCommonMimeTypeWithEncoding_ReturnsFileExtension() {
```
damianszczepanik, line 94: double underscore

src/main/java/net/masterthought/cucumber/json/Embedding.java
```diff
@@ -63,28 +64,35 @@ public String getFileId() {
...
     public String getExtension() {
-        switch (mimeType) {
+        // prepare/ensure switch-case matching
+        String mime = mimeType;
+        // remove subtype's suffix (if existing)
+        if (mime.contains("+")) mime = mime.substring(0, mime.indexOf('+'));
+        // remove subtype's parameter (if existing)
+        if (mime.contains(";")) mime = mime.substring(0, mime.indexOf(';'));
+        // normalise
+        mime = mime.toLowerCase(Locale.ENGLISH).trim();
```
damianszczepanik, line 74: why?

src/test/java/net/masterthought/cucumber/json/EmbeddingTest.java
```diff
@@ -217,6 +230,97 @@ public void getExtension__OnApplicationVNDMsExcelMimeType_ReturnsXls() {
...
+        // given
+        Embedding embedding = new Embedding("application/xslt+xml", "<xsl:stylesheet version=\"1.0\" xmlns:xsl=\"http://www.w3.org/1999/XSL/Transform\" />", "just-some-xslt.weird_extension");
+
+        // when
+        String extension = embedding.getExtension();
+
+        // then
+        assertThat(extension).isEqualTo("xslt");
+    }
+
```
damianszczepanik, line 323: many examples, perfect!

damianszczepanik (changes requested):
Good job, resolve requests and I will be happy to merge it

src/main/java/net/masterthought/cucumber/json/Embedding.java
ghostcity: Dear Damian, at this point we are not referring to the MIME type, but to the name instead. Your example will match the if-condition. However, the extracted 'extension' will not match the file-extension pattern:
```
jshell> String name = "www.google.com/gmail";
name ==> "www.google.com/gmail"

jshell> name.contains(".")
$5 ==> true

jshell> String extension = name.substring(name.lastIndexOf('.') + 1);
extension ==> "com/gmail"

jshell> extension.matches("[a-z0-9]+")
$7 ==> false
```

src/main/java/net/masterthought/cucumber/json/Embedding.java
ghostcity: As specified by RFC-2045 (https://tools.ietf.org/html/rfc2045), "[...] All media type values, subtype values, and parameter names as defined are case-insensitive. [...]".
Thus, there might by upper-case letters and in order to match those cases too, it is converted to lowercase with respect to the english language.

src/main/java/net/masterthought/cucumber/json/Embedding.java
ghostcity: The method now has a JavaDoc documentation.
- https://github.com/damianszczepanik/cucumber-reporting/commit/45261062e90c5fc27294f2e41c2d86fc642bdeeb

src/main/java/net/masterthought/cucumber/json/Embedding.java
ghostcity: Done by https://github.com/damianszczepanik/cucumber-reporting/commit/10c2d22efd21f4b4bd48297763599ed01359b41d

src/test/java/net/masterthought/cucumber/json/EmbeddingTest.java
ghostcity: Actually, I used the similar test name pattern as `getExtension__OnCommonMimeType_ReturnsFileExtension` or `getExtension__OnTextMimeType_ReturnsText`.

There are many pre-existing tests with a name that follows the pattern: "getExtension" + double-underscore + "SomeName" + single-underscore + "ReturnsWhatever".

Do you want me to rename my new test methods only? Or to rename the pre-exisiting tests also?

src/main/java/net/masterthought/cucumber/json/Embedding.java
ghostcity: In addition, the `mime.substring(0, mime.indexOf(';'))` ensures that the MIME-type's parameter is removed. This ensures that, for example, "text/html; charset=UTF-8" will be simplified as "text/html" in order to match the according case in the switch statement.

src/main/java/net/masterthought/cucumber/json/Embedding.java
ghostcity: I'd like to add, that if your example would be the MIME type, the derived extension would be "gmail".
However, this would be an invalid MIME-type because a MIME-type's type cannot contain a ".".
Only the subtype might be written as a tree. (An example is "application/vnd.ms-excel".)

src/test/java/net/masterthought/cucumber/json/EmbeddingTest.java
ghostcity: I tried my best ;-)

src/test/java/net/masterthought/cucumber/json/EmbeddingTest.java
damianszczepanik: interesting approach

damianszczepanik (approved):

damianszczepanik:
Thanks!

ghostcity:
Dear @damianszczepanik , if you agree, I could simplify all the JUnit test by using the `@RunWith(Parameterized)` version.
In result, a single test method would be used to run all similar tests.
Is that OK for you?

damianszczepanik:
Nice idea!

ghostcity:
OK ... #973 is on the way ;-)

Issue #973, opened 2020-11-08
Simplify similar JUnit test (by using the JUnit's Parameterized-Runner)


ghostcity:
As seen by #971, some JUnit tests similar to each other and just differ in the specific test data.
JUnit offers the `@RunWith(Parameterized.class)`-Runner to deal with that in a more compact way.

I'd like to offer such simplification.

PR #979, opened 2020-11-22
Parameterise "StatusCounter" tests (#973)

ghostcity:


src/main/java/net/masterthought/cucumber/json/support/StatusCounter.java
```diff
@@ -76,7 +73,7 @@ public int getValueFor(Status status) {
      * @return sum of all occurrences for all statuses
      */
     public int size() {
-        return size;
+        return counter.values().stream().mapToInt(Integer::valueOf).sum();
```
damianszczepanik, line 76: nice but rather not very efficient if size() is invoked often and report is big

PR #1018, opened 2021-08-16
Shallow cloning on travis

YunLemon:
According to [https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth](url), Travis CI provide a way to shallow clone a repository. This has the obvious benefit of speed, since you only need to download a small number of commits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ cache:
     - ~/.m2
 
 git:
-  depth: false # remove the --depth to support full git blame by Sonarqube
+  depth: 3 # remove the --depth to support full git blame by Sonarqube
 
 env:
   global:


damianszczepanik:
Did you read comment which is on the same line ?

YunLemon:
Very sorry to disturb you. We are doing a research on Travis CI performance problems and hope to get some feedback from the developers. This pull request is automatically created by our tool, and it cannot understand the comments.....
I have never used Sonarqube. The configurations in the .travis.yml file also work for it?
Is full git blame by Sonarqube really important to us?

damianszczepanik:
This change was done intentionally because Sonar does not work properly with configuration you try to apply

PR #1020, opened 2021-08-25
Improve MAVEN build Performance

ChenZhangg:
[Apache Maven Dependency Plugin](https://maven.apache.org/plugins/maven-dependency-plugin/index.html) can be used to find unused dependencies. And I found following list. Maybe we can remove them.
cucumber-reporting
{groupId='com.fasterxml.jackson.datatype', artifactId='jackson-datatype-jsr310'}
{groupId='commons-collections', artifactId='commons-collections'}
{groupId='joda-time', artifactId='joda-time'}
{groupId='commons-configuration', artifactId='commons-configuration'}

According to [Maven parallel test](https://www.baeldung.com/maven-junit-parallel-tests), we can run tests in parallel.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -182,6 +182,9 @@
                         <!-- Integration tests should not be counted by code coverage -->
                         <exclude>${test.integration.pattern}</exclude>
                     </excludes>
+                	<parallel>classes</parallel>
+                	<useUnlimitedThreads>true</useUnlimitedThreads>
+
                 </configuration>
             </plugin>
             <plugin>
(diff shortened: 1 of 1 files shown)


hazendaz:
@damianszczepanik This pull can be closed as it is entirely wrong.  The library suggestions to remove are in fact all used.  The proposed enhancement on surefire would not be very useful, this builds pretty fast already.

PR #1057, opened 2022-07-14
Updated commons-collections dependency to be commons-collections4 to …

astromness:
…remove vulnerability

Updated import in ReportParser.java to use new dependency


astromness:
Fixes Issue-1056

PR #1068, opened 2022-08-16
allow-downloads in html embedding

lucashirsch:
Fix issue where download links are not accessible from within embedded HTML.


damianszczepanik:
Thanks for that

PR #1131, opened 2023-12-22
[test] Use commons text for parts extracted from commons-lang since 3.5

hazendaz:


damianszczepanik:
Can you add context for this change ?

hazendaz:
See https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/text/package-summary.html as straight from javadocs.

As of 3.6+ all the commons text items have been deprecated and were split out to commons-text.  Initially commons text had a hard dependency on commons-lang3 but I believe that is now no longer the case.

For escapeHtml it was further split into 2 pieces.  html4 or html5.  I went with the the lower due to already present issues with libraries well out of date which means it would have been html4 already.  However it could move to html5 on separate pull request.

damianszczepanik:
Can you rebase rest PR to new pom.xml version ?

hazendaz:
> Can you rebase rest PR to new pom.xml version ?

done.

PR #1139, opened 2023-12-27
[java] Try to use more java directly rather than commons*

hazendaz:
java 8 introduced Objects to string which replaces need for commons lang version.

### DIFF
--- a/src/main/java/net/masterthought/cucumber/ReportParser.java
+++ b/src/main/java/net/masterthought/cucumber/ReportParser.java
@@ -9,6 +9,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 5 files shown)


src/main/java/net/masterthought/cucumber/ReportParser.java
```diff
@@ -93,7 +94,7 @@ private Feature[] parseForFeature(String jsonFile) {
             }
             String jsonFileName = extractQualifier(jsonFile);
             Arrays.stream(features).forEach(feature ->
-                    feature.setQualifier(StringUtils.defaultString(configuration.getQualifier(jsonFileName), jsonFileName))
+                    feature.setQualifier(Objects.toString(configuration.getQualifier(jsonFileName), jsonFileName))
```
damianszczepanik, line 97: would have bigger sense if the import of unused class could be deleted or even the whole dependency can go out of pom.xml file

hazendaz: I agree, the issue was it was deprecated, so that was the reason there.  However, to try to show direction better, added a few more adjustments here, renamed the PR, and showing that intent is to eventually drop commons* in general as that is really for java 7 or before rather than java 8+ or as repo was just switched java 11+ now.

hazendaz:
I think by deprecated it was sonarlint telling me that rather than real depreciation as obviously after further looking it was used the same in other areas just not same method type.  And nothing marked those other spots.  One day these tools will be better 😀

PR #1142, opened 2023-12-27
[GHA] Do not run sonarcloud on forks

hazendaz:
Noticed on my fork that sonar cloud was trying to run and of course failing, add line to prevent it from running on anyone but owner of repository.  You may want this partially in other cases such as gh-pages but that one was not failing for me.

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -4,10 +4,11 @@ on:
   push:
     branches:
       - master
-  # do not validate pull requests because SONAR_TOKEN is available only for project owner
+  # Run against master only to not overwhelm sonar given master branch is the base used branch
 
 jobs:
   build:
+    # Do not run sonar on forks because SONAR_TOKEN is available only for project owner
     if: github.repository_owner == 'damianszczepanik'
     runs-on: ubuntu-latest
 
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -8,6 +8,7 @@ on:
 
 jobs:
   build:
+    if: github.repository_owner == 'damianszczepanik'
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/sonarcloud.yml
```diff
@@ -8,6 +8,7 @@ on:
 
 jobs:
   build:
+    if: github.repository_owner == 'damianszczepanik'
```
damianszczepanik, line 11: 1. leave comments in the code why is that
2. with this change there is no need to have branch limitation in line 5-6

damianszczepanik (changes requested):

.github/workflows/sonarcloud.yml
hazendaz: Added variation of the first comment above 'if'.  Adjusted the original.  I think you do still want it only running against master, that is unless you are looking at pull requests.  I can only speak for repos I work on and all simply run on master and nothing else.  If you are looking to more broadly look at those then sure, will remove that.

damianszczepanik: In general - agree with you. My experience is that sonar quite frequently has problem with authentication because of token that is lost somehow. Quite often I did not notice the problem with sonar token and for a long time sonar was not working. Having this on PR (even for myself ) gives opportunity to find such issues earlier :)
Best would be to process sonar against every PR to see tech debt earlier but I was not able to do that :(

damianszczepanik: Also this allows me to run sonar on fix-branch before fix goes to master so I can perform the validation and confirm the fix before it goes to master :)

hazendaz:
Guess more work needed to get sonar part working.

.github/workflows/sonarcloud.yml
hazendaz: @damianszczepanik Per https://docs.sonarsource.com/sonarqube/latest/devops-platform-integration/github-integration/, unless you are a paying customer ($160 a year), its not possible to run on pull requests.  So I've adjusted this a bit to indicate that and point to the documentation.

PR #1143, opened 2023-12-28
Migrate to junit5 jupiter / moderize test library usage / enable all the paramertized tests

hazendaz:
- Fixed one test used in string builder (was looping 'inners' but instead of appending 'inner' was appending 'inners' over and over
- Used open rewrite to run 95% of the migration from junit 4 to junit5.  See https://docs.openrewrite.org/running-recipes/popular-recipe-guides/migrate-from-junit-4-to-junit-5 for more details.
- Fixed the parameterized tests to allow those to run.  This resulted in 345 tests with 2 skips becoming 601 with 27 skips (all those skips are from the original 2).  I didn't look at those specifically to see if they could be fixed.
- Dropped all powermock usage that is junit 4 style on top of mockito as mockito does all that out of the box now.
- Replaced member modification usage of powermock with powermock whitebox which is far easier to read and allowed for rest of obsolete parts of powermock to be removed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -282,20 +282,26 @@
 
         <!-- for testing -->
         <dependency>
-            <groupId>org.junit.vintage</groupId>
-            <artifactId>junit-vintage-engine</artifactId>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
             <version>5.10.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <version>5.10.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>5.8.0</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.powermock</groupId>
             <artifactId>powermock-api-mockito2</artifactId>
-            <version>${powermock.version}</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.powermock</groupId>
-            <artifactId>powermock-module-junit4</artifactId>
             <version>${powermock.version}</version>
             <scope>test</scope>
         </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -282,9 +282,9 @@
 
         <!-- for testing -->
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>4.13.2</version>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <version>5.10.1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -14,6 +14,7 @@
         <maven.compile.encoding>UTF-8</maven.compile.encoding>
 
         <jackson.version>2.16.0</jackson.version>
+        <junit.version>5.10.1</junit.version>
 
         <test.integration.pattern>**/*IntegrationTest*.java</test.integration.pattern>
     </properties>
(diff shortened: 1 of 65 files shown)


pom.xml
```diff
@@ -272,27 +271,27 @@
 
         <!-- for testing -->
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>4.13.2</version>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <version>5.10.1</version>
```
damianszczepanik, line 276: how about moving this version into property, I expect this is the same as in line 282

src/test/java/net/masterthought/cucumber/json/EmbeddingTest.java
```diff
@@ -56,21 +49,15 @@ public static Iterable<Object[]> data() {
...
     public String decodedData;
-
-    @Parameter(3)
     public String fileName;
 
-    @Test
-    public void getMimeType_ReturnsMimeType() {
+    @MethodSource("data")
+    @ParameterizedTest(name = "\"{0}\" with \"{1}\"")
+    public void getMimeType_ReturnsMimeType(String mimeType, String data, String decodedData, String fileName) {
```
damianszczepanik, line 59: to be honest, older convention is more clear to read and understand

src/test/java/net/masterthought/cucumber/json/EmbeddingTest.java
```diff
@@ -133,11 +128,13 @@ public void getExtension_ReturnsFileExtension() {
...
-    @Test
-    public void getExtension_UsesExtensionFromNameWhenMIMETypeIsUnknown() {
+    @MethodSource("data")
+    @ParameterizedTest(name = "\"{0}\" with \"{1}\"")
+    public void getExtension_UsesExtensionFromNameWhenMIMETypeIsUnknown(String mimeType, String data, String decodedData, String fileName) {
+        initEmbeddingTest(mimeType, data, decodedData, fileName);
         // Arrange
-        String mimeType = "unknown/mimetype";
-        String data = "c29tZSBkYXRh";
+        mimeType = "unknown/mimetype";
```
damianszczepanik, line 136: why did you loose type ?

hazendaz: Simple, it was already defined and refused to compile until I did so :)

src/test/java/net/masterthought/cucumber/json/EmbeddingTest.java
hazendaz: I agree, this is harder to get used to but it was automated making it easier.  I've done these by hand and found them complicated in doing so.  I didn't check why the old copies were both ignored but I enabled them and they ran fine so nto sure if there was some issue.  You did have a issue link but it was just about adding the classes and didn't mention why they were set to ignored.

pom.xml
hazendaz: Yup agreed, there are 2 of these.  Once the bots ' renovate or dependabot' run, its better to get one PR than 2.  Will fix that.

src/test/java/net/masterthought/cucumber/json/EmbeddingTest.java
damianszczepanik: Thank you for this upgrade - big effort !

hazendaz:
will fix naming conflict here.

hazendaz:
- Fixed spelling issue on junit version
- Removed the jacoco default as a system property as unnecessary
- Removed few more publics that open rewrite missed

damianszczepanik:
> * Fixed spelling issue on junit version
> 
>     * Removed the jacoco default as a system property as unnecessary
> 
>     * Removed few more publics that open rewrite missed

Great job!

src/test/java/net/masterthought/cucumber/json/EmbeddingWithNameTest.java
```diff
@@ -34,24 +27,16 @@ public static Iterable<Object[]> data() {
...
-    @Parameter(3)
     public String decodedData;
-
-    @Parameter(4)
     public String fileName;
 
-    @Test
-    public void getMimeType_ReturnsMimeType() {
+    @MethodSource("data")
+    @ParameterizedTest(name = "\"{0}\" into \"{2}\"")
```
damianszczepanik, line 37: @hazendaz  after this change tests are still skipped - can you kindly remind me whether was it intentional or tests should be executed and passing again?

hazendaz: Will send PR.  Its the 2 tests that have assumptions to them.  When those assumptions are encountered those tests are skipped.  There are 2 methods like that so total of 12 tests skipped.

hazendaz: Also went ahead and sent some other pull requests I had sitting still.  Nothing critical but I do have some probably 30 commits total I still had not pushed up otherwise so I took a few and pushed up just a while ago.

damianszczepanik: Great, will merge and then perform release!

PR #1147, opened 2024-01-02
Use only slf4j (excluding legacy commons-logging*) and remove logging manipulations during tests as not needed.

hazendaz:
* commons-logging has come back to life but older products are likely long off from supporting and it does more to just redirect to slf4j and log4j2 then any other actual reason for existing.  Since slf4j was already a transient of this project, just make use of slf4j directly and exclude commons logging entirely.  Code itself wasn't using that either really but rather java util logger.

Tests manipulating logs appeared to me as being done to make less output show up on the console, achieved the same thing by simply using slf4j-simple in test scope and changing what gets printed out.

### DIFF
--- a/src/test/java/net/masterthought/cucumber/ReportGenerator.java
+++ b/src/test/java/net/masterthought/cucumber/ReportGenerator.java
@@ -5,8 +5,6 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
 import net.masterthought.cucumber.json.Feature;
 import net.masterthought.cucumber.json.support.StepObject;
@@ -62,7 +60,6 @@ public abstract class ReportGenerator {
     }
 
     protected void setUpWithJson(String... jsonFiles) {
-        Logger.getLogger(ReportParser.class.getName()).setLevel(Level.WARNING);
         initWithJson(jsonFiles);
         createReport();
     }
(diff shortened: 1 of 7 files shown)


src/test/java/net/masterthought/cucumber/ReportGenerator.java
```diff
@@ -62,7 +60,6 @@ public ReportGenerator() {
     }
 
     protected void setUpWithJson(String... jsonFiles) {
-        Logger.getLogger(ReportParser.class.getName()).setLevel(Level.WARNING);
```
damianszczepanik, line 65: the reason of this was to switch off logging mess that was produced every time tests were running - as long as tests pass such logging were not helpful at all

damianszczepanik (changes requested):

damianszczepanik:
Question is: what happens when this change is left unmerged?

hazendaz:
Continues to make users exclude the commons logging library.  We actually force that via a super pom using enforcer plugin.

 With the logging config file I added, you should not incur any additional logging nor cost for the logging as slf4j doesn't log or construct the logger if it's not in the right logging level.  The change to error then blocked all that info which I agree was costly and not useful during tests.  I failed to catch timings before and after but it felt the same and output on console was the same.  I can get some timings on this change.  I believe it's likely a wash.

damianszczepanik:
Can you resolve conflicts?

hazendaz:
yes will catch up here in a few days.  Thanks.

hazendaz:
> Can you resolve conflicts?

Rebased, fixed merge conflict, bumped slf4j to 2.0.11

src/test/java/net/masterthought/cucumber/ReportGenerator.java
hazendaz: With logic in this PR, it will not spam the output with logging and given logging adjustments its faster in general.  Overall though should not really notice much when it comes to the build.

PR #1148, opened 2024-01-02
[pom] Adjust so javadocs are also java 11 now that code base is set to 11

hazendaz:


hazendaz:
This one needs a bit of explaining.  I don't think maven did a great job on naming legacyMode so I have added a comment.  For the past 4 years (since 2019), javadocs have been incredibly broken in case where automatic module naming was added. I don't know that is true here yet but better to protect from that.  In those 4 years there were numerous issues raised, pull requests to fix, etc.  I even raised issues multiple times.  The maven team dismissed the community as wrong and kept carrying on forcing modular that doesn't and hasn't worked out very well at all and maven certainly barely supports it.  It was not until recently someone on their physical team finally ran into this issue and rather than used very well named fix from a contributor PR, they went their own way.  So its legacyMode which doesn't really mean anything (IMO).  What it simply does though is stop trying to erroneously force a project to pretend its modular and not classpath driven.

Now why I added that in addition to setting source to 11.  Well for one, source didn't need set since project itself was set to 11 but given the potential here that the project ends up thinking it is modular and blows up builds, this protects from it happening.  And subsequently we just keep the java 11 flag here separate is it is.

What this will then do for you is remove any javadoc warnings if any for being java 11 build/runtime but stating you wanted java 8 for javadocs.

damianszczepanik (approved):
I wonder: should we have some javadoc check/build for this?

PR #1149, opened 2024-01-02
[pom] Add byte-buddy directly as a java agent

hazendaz:
This resolves newer jdk 19+ that suggests java will block this in the future from dynamically adding.  Therefore this is no longer dynamically added both for lower jdks as well as jdk 17+

How and where byte-buddy is used?
- mockito
- assertj

While the agent is the concern here, we need to make sure both stay in sync as they are well known to fail if out of sync on more recent versions (such as latest here) due to the versions mockito / assertj are using.  That is even if its true, assertj just released new and is up to date now.

For an extremely lengthy discussion on it as well as links/mentions to the JSR involved, see https://github.com/mockito/mockito/issues/3037.  What is being done here is all the way at the bottom and best that can be done for maven currently.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,10 +13,14 @@
         <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
         <maven.compile.encoding>UTF-8</maven.compile.encoding>
 
+        <byte-buddy.version>1.14.11</byte-buddy.version>
         <jackson.version>2.16.0</jackson.version>
         <junit.version>5.10.1</junit.version>
 
         <test.integration.pattern>**/*IntegrationTest*.java</test.integration.pattern>
+
+        <!-- Surefire Setup -->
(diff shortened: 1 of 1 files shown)


damianszczepanik (changes requested):
Can you add some documentation/article to get familiar with this issue/problem?

hazendaz:
Will do.

hazendaz:
@damianszczepanik Added a link to a very lengthy discussion regarding this as it pertains with maven that has been going on for a long time.  It states the JSR where java starts warning as well as fact longer term they will block any dynamic usage, the very last item on that page is how to get around it.  It basically boils down to the consumer must tell java its using an agent until they have an api solution completed (replacement of asm) or they will start blocking such surprising usage.  By stating up front you are using an agent (jacoco for example adds their agent in the build so its somewhat obvious), then it means the consumer of the build is aware an agent is going to manipulate some form of code internally.  In our case during unit tests so the argline works well for that purpose.

PR #1150, opened 2024-01-13
Update libraries again

hazendaz:


pom.xml
```diff
@@ -281,6 +281,11 @@
             <artifactId>owasp-java-html-sanitizer</artifactId>
             <version>20220608.1</version>
         </dependency>
+        <dependency>
```
damianszczepanik, line 284: once the reason of this is mitigated we forget to delete this dependencies - any idea how to address this problem?

damianszczepanik (changes requested):
There is conflict now

pom.xml
hazendaz: Added a comment as to why guava is overridden so its clear we are overriding transient of owasp java html sanitizer until its fixed there.

hazendaz:
Rebased code - mockito updated to 5.10.0, rest is the same.

damianszczepanik (approved):

hazendaz:
rebased again - no changes.

hazendaz:
@damianszczepanik Can we get this merged and also get a release out?

pom.xml
damianszczepanik: Looks like owasp hasn't been released from quite long time thus such problem will happen more and more often

damianszczepanik (approved):

pom.xml
hazendaz: Looked over there https://github.com/OWASP/java-html-sanitizer/commit/3b6cc1b7e6d7992c5b15f87f91dc163de7d35c05, they are removing guava so when they release next we will be able to remove the guava override entirely.  They are going with straight java now.  It appears their activities started back up in January so its promising they likely will be releasing sometime soon as well.  Don't think we need to wait on them as the issue with CVE is fixed by this and they also had multiple tickets asking that fixed before they moved to standard java.

PR #1159, opened 2024-02-03
[GHA] Do not run github-pages unless on main repo

hazendaz:
While the GH Pages upload shows a warning, it is also running this needlessly on forks to that point and shows green.  This will prevent it from running at all so job appropriately shows disabled on forks.


damianszczepanik:
Nice PR - thanks!

PR #1167, opened 2024-03-04
Cleanup pom: compiler and encoding are managed via flags making maintenance easier and other plugins do use those flags too

hazendaz:


damianszczepanik (approved):
I think I have similar issues in other projects

PR #1169, opened 2024-03-04
[java] For java 9+ make sure we use non deprecated new instance

hazendaz:


damianszczepanik (approved):
per https://docs.oracle.com/javase/9/docs/api/java/lang/Class.html#newInstance--

PR #1222, opened 2024-12-26
[GHA] Do not run sonarcloud on forks

hazendaz:
currently this runs on every fork then fails as it should not have been run, this corrects the issue so that it only runs on the upstream repo here as expected.


hazendaz:
added comments to help understanding.

damianszczepanik (approved):

PR #1227, opened 2025-01-11
[tests] Use assert throws instead of try for clarity

hazendaz:
This test performs null pointer exception which probably was not original intent.  This at least makes that clear.


hazendaz:
closing for now as different OS causes different issues, will need to split this test or look at it more as to why its like this.

PR #1240, opened 2025-03-29
refactor: Break cyclic dependency between Configuration and ReportResult

hsbzmn7:
## Problem
DesigniteJava analysis revealed cyclic dependency between:
- `Configuration`
- `ReportResult`

This caused:
- High coupling between reporting logic and configuration
- Difficulty in testing components in isolation

## Solution
- Removed `Configuration` parameter from `ReportResult` constructor
- Pass specific required values (`SortingMethod`, `ReducingMethod[]`)
- Maintained backward compatibility through deprecated constructor


src/main/java/net/masterthought/cucumber/ReportResult.java
```diff
@@ -39,11 +41,23 @@ public class ReportResult {
     private final OverviewReport featuresReport = new OverviewReport();
     private final OverviewReport tagsReport = new OverviewReport();
 
+    // Original constructor for backward compatibility
     public ReportResult(List<Feature> features, Configuration configuration) {
+        this(features,
+                configuration.getSortingMethod(),
+                configuration.getReducingMethods().toArray(new ReducingMethod[0]), // Now correct
+                configuration);
```
damianszczepanik, line 49: Looking at this parameter can you elaborate on what you have wrote, please: 
>> Removed Configuration parameter from ReportResult constructor

damianszczepanik (changes requested):

src/main/java/net/masterthought/cucumber/ReportResult.java
```diff
@@ -39,11 +41,23 @@ public class ReportResult {
...
+    }
+
+    // New constructor
```
damianszczepanik, line 52: useless comment

